PR #2, opened 2014-11-09
Provide ability to pass an R libpath when supporting libraries are not located in standard R paths

chapmanb:
Eric;
This is another bcbio integration request. We put the R libraries in non-standard paths so folks can install without needing sudo permissions. This pull request adds a `--rlibpath` option which will set the given path as the libPath to load libraries. Without the option specified it has the same behavior as previously. Thanks much

### DIFF
--- a/cnvlib/commands.py
+++ b/cnvlib/commands.py
@@ -95,9 +95,9 @@ def _cmd_batch(args):
     pool = pick_pool(args.processes)
     for bam in args.bam_files:
         pool.apply_async(batch_run_sample,
-                        (bam, args.targets, args.antitargets, args.reference,
-                        args.output_dir, args.male_normal, args.scatter,
-                        args.diagram))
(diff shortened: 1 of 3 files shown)


jgarces02:
Yes, this request is totally needed! Is it already implemented? How can we used it? I can't find the instructions to use `--rlibpath`.

Thanks in advance.

chapmanb:
Thanks for the question, CNVkit now respects the settings on your PATH and will use the first Rscript available there. When I run I do `export PATH=/path/to/R/bin:$PATH && cnvkit.py ...` to point at the Rscript with the libraries we want to use. Alternatively you can use `--rscriptpath` which was added to replace this (https://github.com/etal/cnvkit/issues/322). Hope this helps.

jgarces02:
Perfect, with `--rlibpath` (in the `cnvkit.py batch` execution) works perfectly!

Thanks a lot!

Issue #10, opened 2015-03-27
LOH VCF file


pascalg:
The LOH command is not working for me, it doesn't seem to like my VarScan VCF file. Any ideas?

Thanks

```
cnvkit.py scatter Sample.cnr -s Sample.cns -v VarScan.snp.Germline.vcf
Traceback (most recent call last):
  File "/usr/local/bin/cnvkit.py", line 4, in <module>
    __import__('pkg_resources').run_script('CNVkit==0.3.5-dev', 'cnvkit.py')
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 534, in run_script
    if req in processed:
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 1445, in run_script
    
  File "/usr/local/lib/python2.7/dist-packages/CNVkit-0.3.5_dev-py2.7.egg/EGG-INFO/scripts/cnvkit.py", line 9, in <module>
    
  File "build/bdist.linux-x86_64/egg/cnvlib/commands.py", line 714, in _cmd_scatter
    
  File "build/bdist.linux-x86_64/egg/cnvlib/commands.py", line 734, in do_scatter
    
  File "build/bdist.linux-x86_64/egg/cnvlib/ngfrills/vcf.py", line 13, in load_vcf
    # 'format': list of keys (strings)
  File "build/bdist.linux-x86_64/egg/cnvlib/ngfrills/vcf.py", line 33, in filter_vcf_lines
    # When instantiated, a reference can be passed to the VCF class.  This may be any class that supports a
KeyError: 'AF'
```

etal:
It looks like the "AF" (allele frequency, e.g. 0.5 or 1) INFO field is not emitted by VarScan, so I'll have to modify this part of CNVkit to handle it.

Could you post a couple of lines from your VarScan VCF file here? I don't need the whole thing, just a few lines and ideally the header too, if you can.

pascalg:
Thanks, here it is:

```
##fileformat=VCFv4.1
##source=VarScan2
##INFO=<ID=DP,Number=1,Type=Integer,Description="Total depth of quality bases">
##INFO=<ID=SOMATIC,Number=0,Type=Flag,Description="Indicates if record is a somatic mutation">
##INFO=<ID=SS,Number=1,Type=String,Description="Somatic status of variant (0=Reference,1=Germline,2=Somatic,3=LOH, or 5=Unknown)">
[...]

PR #11, opened 2015-03-28
loh plotting: support more callers and multi-sample VCFs

chapmanb:
Eric;
I've been adding all of the awesome CNVkit plotting into bcbio and this contains some small fixes to generalize the LOH plotting to handle more variant callers and multi-sample VCFs:

- Adds support for AF and AD calculation for MuTect, VarScan and FreeBayes
- Enables specifying a sample name to use for LOH calculation instead of first sample.

Thanks much.


etal:
Fantastic. Look like this should resolve issue #10 too. Thanks!

Issue #12, opened 2015-03-30
AssertionError in "diagram" with WGS-level gene density


mjafin:
Hi there,
Thanks for cnvkit, what a great tool!
I've come across this error:

```
Relative log2 coverage of X chromosome: -0.820871 (assuming male)
Traceback (most recent call last):
 File "/group/ngs/src/bcbio-nextgen/latest-devel/rhel6-x64/anaconda/bin/cnvkit.py", line 9, in <module>
   args.func(args)
 File "/group/ngs/src/bcbio-nextgen/latest-devel/rhel6-x64/anaconda/lib/python2.7/site-packages/cnvlib/commands.py", line 685, in _cmd_diagram
   args.output, args.male_reference)
 File "/group/ngs/src/bcbio-nextgen/latest-devel/rhel6-x64/anaconda/lib/python2.7/site-packages/cnvlib/diagram.py", line 48, in create_diagram
   probe_rows = probe_rows.squash_genes()
 File "/group/ngs/src/bcbio-nextgen/latest-devel/rhel6-x64/anaconda/lib/python2.7/site-packages/cnvlib/cnarray.py", line 532, in squash_genes
   outrows.append(squash_rows(name, subarr))
 File "/group/ngs/src/bcbio-nextgen/latest-devel/rhel6-x64/anaconda/lib/python2.7/site-packages/cnvlib/cnarray.py", line 513, in squash_rows
   chrom = core.check_unique(rows['chromosome'], 'chromosome')
 File "/group/ngs/src/bcbio-nextgen/latest-devel/rhel6-x64/anaconda/lib/python2.7/site-packages/cnvlib/core.py", line 154, in check_unique
   % (title, ' '.join(map(str, sorted(its)))))
AssertionError: Inconsistent chromosome keys: chr11_gl000202_random chr17_ctg5_hap1 chr17_gl000203_random
```

I think the error is caused by the gene column containing a dot ( `.` ) as these contigs have no annotated genes.

mjafin:
There's another issue I could use help with in plotting,:
```
Traceback (most recent call last):
  File "/group/ngs/src/bcbio-nextgen/latest-devel/rhel6-x64/anaconda/bin/cnvkit.py", line 9, in <module>
    args.func(args)
  File "/group/ngs/src/bcbio-nextgen/latest-devel/rhel6-x64/anaconda/lib/python2.7/site-packages/cnvlib/commands.py", line 685, in _cmd_diagram
[...]

PR #14, opened 2015-03-30
Ignore bedtools produced '.' specifiers for gene names

chapmanb:
Eric;
This is a small fix for etal/cnvkit#12 to skip over empty gene annotations with '.' as produced by bedtools.


etal:
Thanks! I hadn't encountered that issue myself.

PR #21, opened 2015-04-08
Add option to scatter to plot a BED file of regions

roryk:
Hi Eric,

Thanks for making cnvkit, it rocks. I added an option to scatter to allow a BED file of regions to be specified, and make one multipage PDF of all of the regions together for each region. This is pretty helpful for when there's a bunch of small regions you are interested in looking at all in one go. What do you think about that?


etal:
This is a great idea, thanks! I've been using the shell and pdfunite to accomplish the same until now. Thanks also for the test suite addition.

I'm considering consolidating the "-c" and "-r" options so that "-c chr12" and "-c chr12:345-678" both work. Then I'd change your new "-R" argument to either "-b" (like BED or batch) or "-l" ("list", like samtools mpileup) to avoid the uppercase. Would that throw off your game or cause any workflow problems you can foresee?

roryk:
Hi Eric,

Thanks for the fast response. Consolidating those options sounds like a good idea to me-- I felt bad introducing a lone capital letter but b and r were taken. :)

PR #22, opened 2015-04-11
VCF parsing: handle more AO cases with empty and mulit-allele inputs

chapmanb:
Eric;
This is a small fix for VCF parsing to handle more AO cases we ran into when doing more testing: empty and multi-allelic. Hopefully this catches everything now.


etal:
Looks great, thanks!

Issue #18, opened 2015-04-01
scatter: Identify and highlight LOH segments in plot


etal:
The `loh` plot used to try to identify whole chromosomes with loss of heterozygosity (LOH) using an ill-conceived Mann-Whitney test. This would fail if multiple chromosomes had LOH (skewing the background distribution), or if the LOH region was much smaller than a chromosome (i.e. covering too few SNPs).

This issue has two parts:
- If segments are provided to the `scatter` command, group variants in the given VCF by segment. (Mostly done.)
- Within a segment (or whole chromosome, if segments are not given), use a statistical test to determine if SNP allele frequencies in the region differ from those with neutral-frequency. See PSCBS paper and R package since @HenrikBengtsson already put the necessary thought into this.

See also: #34

PR #23, opened 2015-04-15
pure code clarity commit

roryk:
Just deleted some dead code and fixed a PEP8y formatting issue.


etal:
Thanks, I agree with this. The LOH thing is addressed in #18 and the plot aesthetics issue (left-side ticks facing out, not in) clearly requires code different than what's commented out -- though it would be nice to fix both of those eventually.

PR #24, opened 2015-04-16
diagram plots: sanity check features to ensure none expand over chromosome boundaries

chapmanb:
Eric;
We ran into a few cases when testing where the plotted features went over the ends of chromosomes causing plotting failures in Bio.Graphics. This removes any features with those issues to avoid the problem.

### DIFF
--- a/cnvlib/diagram.py
+++ b/cnvlib/diagram.py
@@ -44,6 +44,7 @@ def create_diagram(probe_pset, seg_pset, threshold, outfname, male_reference):
     no_name = ('Background', 'CGH', '-', '.')
     strand = 1 if do_both else None
     probe_rows = core.shift_xx(probe_pset, male_reference)
+    chrom_sizes = plots.chromosome_sizes(probe_pset)
     if not is_seg:
         probe_rows = probe_rows.squash_genes()
     for row in probe_rows:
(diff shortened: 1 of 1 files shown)


etal:
I dig it. Does this also fix #12?

chapmanb:
Thanks for merging this is so fast. I don't think this tackles #12. I actually worked on that upstream of cnvkit and remove the smaller chromosomes prior to running the diagram plots (since you don't really want those anyhow):

https://github.com/chapmanb/bcbio-nextgen/blob/0170a173a1d6a1699b4a897e78fdd17dd3395636/bcbio/structural/cnvkit.py#L194

I tried to generalize it to cluster by size to determine the "small" threshold instead of explicitly removing specific chromosome but not sure how awesome that code is.

etal:
Hmm, clustering is hard. I tried your code on hg19 and found that using k-means once with k=3 also finds 24 non-little chromosomes, the same as two iterations of k=2. Better than filtering on chromosome names, in any case. I'll see if I can work it into CNVkit.

chapmanb:
Good idea. I thought it would magically separate big/tiny with a single pass of k=2 but had no luck with that, so did the ugly 2 pass approach. There might be better ways to get the separation and say remove anything too tiny, so happy for any ideas. Thanks for looking at this.

PR #27, opened 2015-04-24
Use Agg backend to avoid DISPLAY errors when running on remote servers.

chapmanb:
Eric;
This is a small fix that avoids Invalid DISPLAY errors when running cnvkit plotting on remote servers without X.

http://stackoverflow.com/questions/4931376/generating-matplotlib-graphs-without-a-running-x-server
http://stackoverflow.com/questions/2801882/generating-a-png-with-matplotlib-when-display-is-undefined


etal:
I added a check for $DISPLAY, and I think this ought to work on remote Linux servers as well as Macs. Could you try it and let me know if it works for you?

PR #30, opened 2015-05-07
Fixes for edge cases encountered in testing: empty and one line CNA files and problem VCFs

chapmanb:
Eric;
This pull request has a few fixes encountered when running on especially stubborn real life cases:

- Empty CNA files would return an empty list, which resulted in errors downstream for objects expecting a CNA, not a list. It now returns an empty CNA instead.
- Loading CNAs with a single items resulted in a data attribute with different dimensions than a full multi-item file.
- I found some VCF cases which just don't have alternative allele attribute so we warn and skip those.

Thanks much.


etal:
Awesome, thanks for this.

PR #42, opened 2015-08-03
Anti-target and VCF export fixes during testing

chapmanb:
Fixes a few issues found during initial VCF export integration:

- The new antitarget logic can create duplicate regions, resulting
  in a pandas downstream error when attempting to index on the data
  frame.
- Allow specification of the sample ID to write in VCF output.
- Remove a few confusing header fileds in the VCF
  (wrong reference/file dates).


etal:
Looks great, thanks!

PR #50, opened 2015-09-25
Add .coveragerc

stevepeak:
> from your Codecov support request

I believe this will collect all the coverage files missing in your reports

Run `coverage report` in your local setup to confirm the files. 

http://coverage.readthedocs.org/en/latest/config.html


etal:
I think this is working now, thanks!

PR #53, opened 2015-10-02
Handle cns files with missing probe counts

chapmanb:
Eric;
We ended up with a .cns file having a '_' for `probes` which
caused issues on export to VCF since pandas converted the `probes`
column to a string instead of the expected integer. This is a band-aid
that skips rows without probes and assures other rows are integers
before feeding into string formatting. I'm not sure if other upstream
changes are more appropriate but this works for us since we're
passing VCF call files along.

This also bumps the version to 0.7.2.dev0 since .dev0 versions sort
before full versions (0.7.1.dev0 < 0.7.1).

Thanks much.

### DIFF
--- a/cnvlib/_version.py
+++ b/cnvlib/_version.py
@@ -1,1 +1,1 @@
-__version__ = "0.7.1.dev0"
+__version__ = "0.7.2.dev0"
--- a/cnvlib/export.py
+++ b/cnvlib/export.py
@@ -268,12 +268,12 @@ def segments2vcf(segments, ploidy, is_reference_male, is_sample_female):
     # TODO be more clever about this
     for (_idx, out_row), (_idx, abs_row) in zip(out_dframe.iterrows(),
                                                 abs_dframe.iterrows()):
-        if out_row["ncopies"] == abs_row["expect"]:
+        if out_row["ncopies"] == abs_row["expect"] or out_row["probes"] == "_":
             # Skip regions of neutral copy number
             continue  # or "CNV" for subclonal?
 
         if out_row["ncopies"] > abs_row["expect"]:
-            genotype = "0/1:0:%d:%g" % (out_row["ncopies"], out_row["probes"])
+            genotype = "0/1:0:%d:%g" % (out_row["ncopies"], int(out_row["probes"]))
         elif out_row["ncopies"] < abs_row["expect"]:
             # TODO XXX handle non-diploid ploidies, haploid chroms
             if out_row["ncopies"] == 0:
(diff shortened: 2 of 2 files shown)


cnvlib/export.py
```diff
@@ -268,12 +268,12 @@ def segments2vcf(segments, ploidy, is_reference_male, is_sample_female):
     # TODO be more clever about this
     for (_idx, out_row), (_idx, abs_row) in zip(out_dframe.iterrows(),
                                                 abs_dframe.iterrows()):
-        if out_row["ncopies"] == abs_row["expect"]:
+        if out_row["ncopies"] == abs_row["expect"] or out_row["probes"] == "_":
```
etal: What if there's a dot, empty string, or NaN here instead? Keeping the same semantics, I'm inclined to do `or not isdigit(str(out_row["probes"]))`, maybe.

cnvlib/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.7.1.dev0"
+__version__ = "0.7.2.dev0"
```
etal: Is there an incantation for "later than the last stable version"? Some tools don't agree with the "+" suffix Biopython uses.

etal:
I agree with the motivation. Is there a reason to favor `_` over `.`, `-`, or `NA`? I've created files with `0` probes to indicate "dunno" and I can imagine leaving the field blank is also reasonable.

cnvlib/_version.py
chapmanb: I don't know of that on. I always use a "pre next version" tag. I use alpha (0.7.2a0) but I think dev works fine too.

cnvlib/export.py
chapmanb: Sounds reasonable to me. cnvkit output the cns file with the underscore, so I thought that's what you used but happy to have this more general. Do you want me to add that fix?

chapmanb:
Eric -- thanks. If underscore isn't a convention in CNVkit it's great to generalize this. This was more of a band-aid to fix an issue as I didn't track down the original source of the underscore column. Thanks much.

cnvlib/export.py
etal: Sure, if you don't mind. Then I'll just merge this pull request.

chapmanb:
Eric -- all set with the generalize checks for non-numerical probes. Thanks again.

etal:
Thanks! Lemme know if you find out where the .cns file came from and what problem the _ was meant to address, and maybe I can find another solution that doesn't break parsing.

etal:
I figured it out. During segmentation via R (I noticed this for "flasso"), the dataframe columns could get reordered. When all of a chromosome's bins are filtered out for poor coverage, a placeholder segment is generated, but the dummy gene name "_" was mistakenly plugged into the "probes" column because of this reordering. Fixed in 2328ec94bbe7598015d82bc3a1d9096b58c7f867 so it shouldn't happen any more. The low-coverage filter is now optional too (653dc07d712f22c0d5b409e5063d82769247b018) because it was burning people with well-mapped germline samples.

I'll leave this pull request's change in place, at least for a while, since some of these goofy .cns files now exist in the wild.

chapmanb:
Nice detective work. Thanks for fixing the actual problem. Much appreciated.

PR #66, opened 2016-01-01
Added R channel to conda command

kyleabeauchamp:
As of recently (https://github.com/bioconda/bioconda-recipes/pull/523), the cnvkit recipe in bioconda will install the R dependencies necessary for segmentation.  For this to work, however, we need to add the `r` channel to the conda install command.


etal:
Good eye, thanks for the fix.

PR #78, opened 2016-02-17
Fix coverage docs in pipeline.rst

kyleabeauchamp:
AFAIK (https://github.com/etal/cnvkit/blob/db70037e427c51b00a79c7962e33b9efbe2241c1/cnvlib/commands.py#L453) there is no `-p` option for coverage.  Furthermore, I believe the default is to count via pileup and use midpoint counting only if `--count` is specified.  Correct me if I am wrong here.


etal:
Thanks!

PR #89, opened 2016-03-20
Fixing tuple bug when using fused lasso

pcingola:
Probably not the most elegant solution, but it seems to work OK now.


etal:
Thanks, this one was due to a switch from iterating over the underlying pandas dataframe using `iterrows`, which returns a dict-like Series per row, to `itertuples`, which returns namedtuples that can be indexed by position or by name, as an attribute. I just pushed a change to do it the latter way.

PR #91, opened 2016-03-29
Add VCF fields info field

pcingola:
Currently "cnvkit export" does not report FOLD_CHANGE in the VCf file. Added 
some VCF fields that might be useful (FOLD_CHANGE, PROBES, etc. directly 
copied from *.cns).


etal:
Thanks, I've cherry-picked the commit with these changes. Are these new fields standard or common, or have you seen them in VCF files generated by other programs?

pcingola:
I've seen other programs create fold change information. These may not be 
the same INFO field names though. Independently of the field name, having fold change 
info is useful for downstream processing. Thank you!

Issue #3, opened 2015-03-05
Not working in python 3?


dawe:
I've installed cnvkit using pip, at first run it raises a SyntaxError:

```
$ cnvkit.py -h
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.4/bin/cnvkit.py", line 5, in <module>
    from cnvlib import commands
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/cnvlib/__init__.py", line 1, in <module>
    from .cnarray import CopyNumArray as _CNA
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/cnvlib/cnarray.py", line 11, in <module>
    from . import core, metrics, ngfrills, smoothing
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/cnvlib/core.py", line 9, in <module>
    from .ngfrills import echo, safe_write
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/cnvlib/ngfrills.py", line 407
    except OSError, exc:
                  ^
SyntaxError: invalid syntax
```

After correction of entries in python3 syntax (i.e. `OSError, exc` becomes `OSError as exc`) I have

```
$ cnvkit.py 
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.4/bin/cnvkit.py", line 9, in <module>
    args.func(args)
AttributeError: 'Namespace' object has no attribute 'func'
```

etal:
CNVkit isn't currently designed to work on Python 3, but thanks for bringing this up. If you have a changeset that improves Python 3 compatibility, I'll see if I can merge it.

It looks like argparse has changed in some crucial way between Pythons 2.7 and 3.x, and that will require some futher investigation.

mpschr:
Hi - I have created a pull request with python3 compatibility
 #100

mpschr:
I guess this can be closed @etal

PR #100, opened 2016-05-03
Python3 compatibility and option "ignore_gt_field"

mpschr:
Hi

I have gotten the code to work with python3. I have been able to use the functions `do_call`, `VariantArray.read_vcf` and `CopyNumArray.read` in my python3 jupyter notebook.

I have run the unittests for both python2 and python3 and I think everything works fine by now. I just wanted to let you know in case you are interested in making a pull request - particularly for the python3 compatibility which I think may serve well for more people ;)

Unrelated,  I have added an option to `VariantArray.read_vcf` called `ignore_gt_field` which lets me ignore the GT (genotype) field of the `vcf` file since I have hand-selected the SNPs for ExAC positions and being heterozygous in the matched normal. The original problem is actually, that I have generated the vcf files with `samtools mpileup` and thus I don't even have the GT field in my `.vcf` file. I guess this may be a problem more people may encounter.

Let me know if there are any more questions/doubts or whatsoever

Best,
Michael

PR #101, opened 2016-05-03
Python3 compatibility

mpschr:
As with the old pull request (#100) unit test seems to work.

Hopefully the future dependency in the setup.py solves the Travis CI problem


mpschr:
fix #3

PR #102, opened 2016-05-03
Ignore GT (genotype) field if necessary

mpschr:
I have added an option to `VariantArray.read_vcf` called `ignore_gt_field` which lets me ignore the GT (genotype) field of the `vcf` file since I have hand-selected the SNPs for ExAC positions and being heterozygous in the matched normal. The original problem is actually, that I have generated the vcf files with `samtools mpileup` and thus I don't even have the GT field in my `.vcf` file. I guess this may be a problem more people may encounter.

### DIFF
--- a/cnvlib/vary.py
+++ b/cnvlib/vary.py
@@ -78,8 +78,12 @@ class VariantArray(gary.GenomicArray):
         rows = _parse_records(vcf_reader, sample_id, normal_id, skip_reject, ignore_gt_field)
         table = pd.DataFrame.from_records(rows, columns=columns)
         table["alt_freq"] = table["alt_count"] / table["depth"]
+        if ignore_gt_field:
+                table["zygosity"] = (2 * table["alt_freq"]).round() / 2
         if normal_id:
             table["n_alt_freq"] = table["n_alt_count"] / table["n_depth"]
+            if ignore_gt_field:
+                table["n_zygosity"] = (2 * table["n_alt_freq"]).round() / 2
         # Filter out records as requested
         cnt_depth = cnt_hom = cnt_som = 0
         if min_depth:
--- a/cnvlib/vary.py
+++ b/cnvlib/vary.py
@@ -58,7 +58,7 @@ class VariantArray(gary.GenomicArray):
 
     @classmethod
     def read_vcf(cls, infile, sample_id=None, normal_id=None, min_depth=None,
-                 skip_hom=False, skip_reject=False, skip_somatic=False):
+                 skip_hom=False, skip_reject=False, skip_somatic=False, ignore_gt_field=False):
         """Parse SNV coordinates from a VCF file into a VariantArray."""
         if isinstance(infile, basestring):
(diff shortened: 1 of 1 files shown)


cnvlib/vary.py
```diff
@@ -252,19 +256,19 @@ def _parse_records(vcf_reader, sample_id, normal_id, skip_reject):
...
         depth = alt_count = None
     if sample.is_het:
         zygosity = 0.5
-    elif sample.gt_type == 0:
+    elif not ignore_gt and sample.gt_type == 0:
         zygosity = 0.0
     else:
         zygosity = 1.0
-    alt_count = (_get_alt_count(sample) if sample.gt_type else 0.0)
+    alt_count = _get_alt_count(sample) if ignore_gt or sample.gt_type else 0.0
```
etal: If we just remove the original `if sample.gt_type else 0.0`, and the user can set the `zygosity` values to 0.5 after parsing, does that fix the problem without introducing a new option?

etal:
The procedure I'm suggesting is to call `read_vcf` keeping the default `skip_hom=False` keyword argument, so all records are retained regardless of GT status. Then the caller can afterward set `variants.zygosity = 0.5` or some thresholding (I've seen .16/.84 suggested as a good cutoff) to apply the desired genotype calls.

Then for the command line I can add another option for ignoring genotypes (and perhaps handling other missing-data issues in the VCF) that would do all of the above. Would that solve the original problem you found?

cnvlib/vary.py
mpschr: It should work, but I would rather not set the zygosity and derive it from allele frequency later on

mpschr:
I think this way would work for my case where I don't have the GT field in the .vcf. I am not sure though how it would behave with the `skip_hom=True` and GT is not field is available.

So what would work for all cases would be something as follows:
 * `read_variants` without discarding homozygous
 * add a function `discard_homozygous` which can be called by the parser (if `skip_home=True`) or user via API if wanted (e.g. when creating the `VariantArray` from a `DataFrame`).
  * if no GT field is available in the upon reading the vcf or the `.data` table does not contain the `zygosity` field emit a info/warning like: `deriving zygosity from allele frequency` and do that.

I have to say, that in the end I am happy with any solution that does not discard my variants because there is a `LOH` in my tumor sample (and no normal sample VAF supplied).

etal:
I've changed the VCF processing to behave more or less according to your last comment here -- reading a VCF file does not discard homozygous or no-GT variants, and instead the `scatter` and `export` commands use the new VariantArray method `heterozygous()` to select heterozygous germline SNPS. If no variants are parsed as heterozygous, `heterozygous()` keeps everything. If no depths (DP) are available, `--min-variant-depth=0` can be used to keep everything. For the API, VariantArray.heterozygous() has the optional keyword arguments `min_freq` and `max_freq` that can be used to select variants with allele frequencies between those two values.

mpschr:
Hi, sounds great! Feel free to close the pull request

PR #104, opened 2016-05-04
Empty gene names throw error for scatter plot

mpschr:
There was a problem when a segment has no gene name (therefore np.NaN value) and the code wanted to perform a `.split(',')` on the gene name.

solution: `.fillna('')`

### DIFF
--- a/cnvlib/plots.py
+++ b/cnvlib/plots.py
@@ -489,7 +489,7 @@ def gene_coords_by_name(probes, names):
     """
     # Create an index of gene names
     gene_index = collections.defaultdict(set)
-    for i, gene in enumerate(probes['gene']):
+    for i, gene in enumerate(probes['gene'].fillna('')):
         for gene_name in gene.split(','):
             if gene_name in names:
                 gene_index[gene_name].add(i)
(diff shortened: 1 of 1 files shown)


mpschr:
HI @etal , is there something I should do for the checks not to fail? I don't fully understand how it works. ;)

etal:
1. How did the gene names end up missing to begin with? Did this happen when using the `cnvlib` library, or did you get a .cnr file that was parsed as having "NaN" gene names? If it's the latter, I'd prefer to catch this weird case earlier in processing, e.g. during instantiation, so that the other commands don't have to worry about it either.
2. Don't worry, your code is fine. It's passing on Python 2.7, but I haven't figured out how to get Travis CI to install pysam successfully on Python 3.

mpschr:
Hi

I think it's just that there are empty values in the tabular file since not every bin has a gene name assigned. Empty values are being filled with `numpy.NaN` by default, altough the rest of the values are `str`

You are right that it makes more sense to replace `numpy.NaN` with `""` (empty `str`) upon reading.

Best,

etal:
I applied your suggestion at the point in the code where the "gene" column is created. Thanks!

PR #108, opened 2016-05-13
Fix genes not split correctly for gainloss report

### LINKED ISSUE #107: Gainloss report does not fit description of documentation
Here the documentation description: 

> If segments are given, the log2 ratio value reported for each gene will be the value of the segment covering the gene. Where more than one segment overlaps the gene, i.e. if the gene contains a breakpoint, each segment’s value will be reported as a separate row for the same gene. If a large-scale CNA covers multiple genes, each of those genes will be listed individually.

My output looks like this:

*called.cns*:
```9       21960001        22860000        many,genes,CDKN2A,CDKN2B,RP11-145E5.5        -2.656  45      0.680778        0       0       0```

*gainloss*:
```
RP11-145E5.5,C9orf53,CDKN2A        9       21960001        21980000        -2.656    1
RP11-145E5.5,CDKN2A,CDKN2B-AS1     9       21980001        22000000        -2.656    1
RP11-145E5.5,CDKN2B-AS1,CDKN2B     9       22000001        22020000        -2.656    1
RP11-145E5.5,CDKN2B-AS1            9       22020001        22040000        -2.656    1
```

But according the description I'd expect something like this. One gene per row and the probe number according to the segmented file

```
RP11-145E5.5     9       21960001        22040000        -2.656    45
CDKN2A           9       21960001        22000000        -2.656    45
CDKN2B-AS1       9       21980001        22040000        -2.656    45
CDKN2B           9       21980001        22040000        -2.656    45
```

The question is - if I am right with my assumption I can upload a pull request since I am adapting it now

Let me know

### PR DESCRIPTION
**Pull request by mpschr**
This commit should fix #107 

a couple of details:
 * `gary.py` now has a function called `_get_gene_map` which returns an OrderedDict of the gene names (all split) and their respective indices in the data. The order of the map is the same as of occurrence

* if `gainloss` is performed with a `called.cns` the additional columns such as `cn`, `baf` are being maintained as the `segment` is being updated with `gene`, `start` and `end`, whereas all other values are being maintaine - including `log2` and `probes` are from the segmented information. Particularly the `probes` are important to come from the segmented file, since filtering is being done upon that value.

* There is a pandas bug which affects the function `gary.by_chromosome` upon grouping of a categorical. I had to inject a `astype(str)` - otherwise it will not work. Can be removed in the future.

### DIFF
--- a/cnvlib/reports.py
+++ b/cnvlib/reports.py
@@ -8,6 +8,8 @@ import collections
 import math
 import sys
 
+import numpy as np
+
 from . import metrics, params
 
 iteritems = (dict.iteritems if sys.version_info[0] < 3 else dict.items)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cnvlib/gary.py [hunk 1]
```diff
@@ -495,3 +501,22 @@ def write(self, outfile=None):
         with ngfrills.safe_write(outfile or sys.stdout) as handle:
             self.data.to_csv(handle, index=False, sep='\t', float_format='%.6g')
 
+    def _get_gene_map(self) -> OrderedDict:
```
etal: Type annotations are deep in Python 3 territory; Python 2.7 needs this information in the docstring only.

cnvlib/gary.py
```diff
@@ -204,6 +205,11 @@ def autosomes(self, also=()):
 
     def by_chromosome(self):
         """Iterate over bins grouped by chromosome name."""
+
+        # necessary because of pandas bug which says:
```
etal: Does the pandas bug correspond to an issue in the upstream repo (pydata/pandas) that we can link to?

cnvlib/gary.py
mpschr: You're absolutely right - fixing it!

cnvlib/gary.py
mpschr: I think these two below are tightly related - but I had no time yet to hunt down the bug in detail and confirm how they are related.

* pydata/pandas#10505
* pydata/pandas#10508

mpschr: So @etal 

I have found the "bug causing element" and reported the following bug here:  pydata/pandas#13179

cnvlib/reports.py
```diff
@@ -86,9 +86,9 @@ def gainloss_by_segment(probes, segments, threshold, skip_low=False):
     """
     for segment, subprobes in probes.by_ranges(segments):
         if abs(segment.log2) >= threshold:
-            for (gene, chrom, start, end, _coverage, nprobes
-                ) in group_by_genes(subprobes, skip_low):
-                yield (gene, chrom, start, end, segment.log2, nprobes)
+            for (gene, chrom, start, end, _coverage, nprobes) in \
+                    group_by_genes(subprobes, skip_low):
+                yield segment._replace(gene=gene, start=start, end=end)
```
etal: This is causing the unit test to fail. The ordering of columns in the `gainloss` output is different, so keeping the extra columns is going to need a different strategy. (It also needs to be done the same with and without the `segments` input.) E.g.:

     columns = ['Gene', 'Chrom.', 'Start', 'End', 'Log2Ratio', 'Probes']
     extra_column_names = cnarr.data.columns.tolist()[len(cnarr._required_columns):]
     columns.extend(extra_column_names)

mpschr: Hi, sorry - I had forgotten the unittests. Now they work fine.

The problem actually was that I was expecting the extra columns in `gainloss_by_gene` where it makes no sense (Impossible to know what kind of aggregation wanted for extra columns).

I fixed that in the following commit: a46acc0

I also fixed the column order to move `gene` to the front as it was before my commits.

etal: I'd recommend reverting the lines that alter output columns for this pull request, to fix the original bug. Then you can experiment with extending the columns if you like, or file a separate issue and I can do it if you prefer.

mpschr: Hi. Now it is again in the format in which you had it right? 

The only thing that changes is, if used with the `called.cns`, the called colums are being maintained. 
That's all I think.

Cheers

etal:
I've merged your commits and made a couple more on the same topic. Care to try the `master` branch now and see if it works the way you wanted?

PR #120, opened 2016-08-22
scatter: Allow output of non-pdf plots

chapmanb:
Eric;
I'd like to be able to include scatter outputs into HTML reports and this makes it possible to directly produce non-pdf plots. Thanks much for considering the change.


etal:
Rad. Thanks!

Issue #55, opened 2015-10-07
Parallel "batch" sometimes fails to generate output files (.cnn/.cns)


etal:
The `batch` command when run on multiple CPU cores in parallel (`-p 0` or `-p` >1) fails to create all of the expected output files, but doesn't indicate why.

This could be more than one underlying issue:

- If segmentation crashes under certain conditions, it's a silent error and the command simply completes without creating the .cns file. But the .cnr file has been successfully produced by this point.

*Workaround:* Run the `segment` command on the .cnr file to see the error that caused segmentation and therefore `batch` to originally fail/abort. Fix the error if you can, or report the issue here and wait for a timely fix. Then if `segment` can be run successfully, parallel `batch` probably can, too. (Or just finish the job manually by running `segment` on each sample separately.)

- Reported in chapmanb/bcbio-nextgen#915: With `-p` > 1, the `batch` pipeline produces binned coverages and a reference (.cnn files), but fails to create a .cnr file and/or .cns file.

*Workaround:* Run the same command with `-p 1` instead, and the pipeline will successfully run on a single core. 

*Diagnosis:* There's an issue either in CNVkit's use of the multiprocessing library, or the multiprocessing library in Python 2.7 itself has an issue synchronizing on some platforms. It works on Ubuntu 14.04 and later; I'm interested to hear where it doesn't work.

wjaratlerdsiri:
When batch -p 12 stopped with sample.cnr file, I ran below and got errors:

./python2.7 cnvkit.py segment sample.cnr -o sample.cns


Dropped 21 outlier bins:
   chromosome      start        end          gene      log2  weight
0        chr1   91460164   91460431       MIR3675  -6.24330       0
1        chr1  202172548  202172815      RNVU1-17  -3.66398       0
2        chr2  122144558  122144825         COX5B  -3.69414       0
3        chr2  136347229  136347496         COX5B  -5.11515       0
4        chr2  169466973  169467240         COX5B -25.15570       0
5        chr3    1707598    1707865     ITPR1-AS1  -8.33714       0
6        chr3  169770460  169770726       MIR8060  -3.66135       0
7        chr6   49096817   49097084         HUS1B  -4.86744       0
8        chr6   54252019   54252286         HUS1B  -5.99444       0
9        chr7    7679073    7679340  LOC100507642 -25.16300       0
10       chr8   33318966   33319233       DEFB130  -7.19119       0
11       chr8   35248295   35248562       DEFB130  -6.22393       0
[...]

PR #121, opened 2016-08-25
WIP: initial stab at parallelizing coverage

brentp:
This isn't ready, but I want to post early to see if there is interest and to get feedback.

I am using @chapmanb 's bcbio to parallelize across samples but notice that there are many points where cnvkit is running single-threaded.

Here, I have parallelized the call to bedcov on every 5K intervals. It's seeming to work quite nicely on my test dataset.

Do you have interest in including this with the appropriate changes to expose to the command-line?

### DIFF
--- a/cnvlib/coverage.py
+++ b/cnvlib/coverage.py
@@ -5,8 +5,13 @@ from past.builtins import basestring
 
 import logging
 import math
+import os
 import os.path
+import atexit
+import tempfile
 import time
+import gzip
+from concurrent import futures
 
 import pysam
 
@@ -63,18 +68,21 @@ def interval_coverages(bed_fname, bam_fname, by_count, min_mapq):
                          meta_dict={'sample_id': fbase(bam_fname)})
 
 
-def interval_coverages_count(bed_fname, bam_fname, min_mapq):
+def interval_coverages_count(bed_fname, bam_fname, min_mapq, procs=12):
     """Calculate log2 coverages in the BAM file at each interval."""
     bamfile = pysam.Samfile(bam_fname, 'rb')
-    for chrom, subregions in tabio.read_auto(bed_fname).by_chromosome():
-        logging.info("Processing chromosome %s of %s",
-                     chrom, os.path.basename(bam_fname))
-        for _chrom, start, end, gene, in subregions.coords(["gene"]):
-            count, depth = region_depth_count(bamfile, chrom, start, end,
-                                              min_mapq)
-            yield [count,
-                   (chrom, start, end, gene,
-                    math.log(depth, 2) if depth else NULL_LOG2_COVERAGE, depth)]
+    with futures.ProcessPoolExecutor(procs) as pool:
+
+        for chrom, subregions in tabio.read_auto(bed_fname).by_chromosome():
+            logging.info("Processing chromosome %s of %s",
+                         chrom, os.path.basename(bam_fname))
+
+            for count, depth in pool.map(region_depth_count, (
+                                         (bamfile, chrom, start, end, min_mapq)
+                                         for _chrom, start, end, gene in subregions.coords["gene"])):
+                yield [count,
+                       (chrom, start, end, gene,
+                        math.log(depth, 2) if depth else NULL_LOG2_COVERAGE, depth)]
 
 
 def region_depth_count(bamfile, chrom, start, end, min_mapq):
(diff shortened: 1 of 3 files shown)


cnvlib/coverage.py
```diff
@@ -104,15 +112,60 @@ def filter_read(read):
...
+    try:
+        os.unlink(path)
+    except OSError:
+        pass
+
+def to_chunks(bed_fname, chunk_size=5000):
+    """Split the bed-file into chunks for parallelization"""
+
+    k, chunk = 0, 0
+    fd, name = tempfile.mkstemp(suffix=".bed", prefix="tmp.%s." % chunk)
```
etal: Would tempfile.NamedTemporaryFile make this any cleaner? It has an option to delete the temporary file when it goes out of context.

cnvlib/coverage.py
```diff
@@ -104,15 +112,60 @@ def filter_read(read):
...
+    fh = os.fdopen(fd, "w")
+    atexit.register(rm, name)
```
etal: What does this do?

etal:
Fantastic. For exomes and especially WGS, it's a huge help to split `coverage` and possibly `segment` across chunked files. The `batch` command parallelizes across samples, but not at any finer grain.

`batch` uses plain `multiprocessing` and a shared worker pool, so the command line option `-p` can be used to specify the number of processes -- see `cnvlib/parallel.py`. It would be nice for the same parallelization approach to be used across the codebase.  (The current approach has room for improvement, see also #55)

The usage of `concurrent.futures` here looks similar to the `map` and `apply_async` methods in plain multiprocessing; is `futures` better in other ways? E.g. would it be feasible/easy to pass the pool around and use it throughout `batch` as well as in the various steps it calls, while keeping the total number of processes within the specified limit?

I've also though about using `pandas.DataFrame.groupby('chromosome').apply(...)`, which might release the GIL. The BED files in CNVkit are supposed to live mostly in memory as dataframes, so there could be some other pandas tricks to simplify the code here.

cnvlib/coverage.py
```diff
@@ -104,15 +112,61 @@ def filter_read(read):
...
 
 
-def interval_coverages_pileup(bed_fname, bam_fname, min_mapq):
+def rm(path):
+    try:
+        os.unlink(path)
+    except OSError:
+        pass
+
+def to_chunks(bed_fname, chunk_size=5000):
```
etal: The chunk size could be calculated on the fly from the number of processes and lines in the BED file. The BED file can turn into a GenomicArray instance with some bedtools-like features via `tabio.read(bed_fname, "bed")`.

cnvlib/coverage.py
brentp: This is so the the file gets removed even if the process exits unexpectedly.

cnvlib/coverage.py
brentp: Since I'm returning the file name from the funciton, that exits the with scope. To use named-temporary file, we'd have to inline this into the calling function. What do you think?

cnvlib/coverage.py
brentp: On my data, 5000 lines takes about 1 minute per process so it stays nicely parallel. I was thinking that the bigger the chunk, the more risk there is of not keeping the cores busy. Though certainly something like 20,000 should be fine.

I'm happy to use your machinery here if you have something else in mind.

brentp:
I'm told that concurrent futures is the way forward in python 3 as it's in the stdlib and it's back-ported to python 2. I can switch to using multiprocessing if you prefer.

So the questions for you:
+ switch this PR to multiprocessing or stay with concurrent.futures?
+ inline to_chunks so we can use `with NamedTemporaryFile`? [I prefer not to do so on this one]
+ how to set the chunk size for to_chunks?

I don't have strong opinions on any of those; I'll defer to your preference.

Meanwhile, I'll integrate to use the -p command.

p.s. I also so the single-threaded bottle-neck in segment and saw your comments in the code about that. I can add to split by chrom once this one is resolved.

brentp:
BTW, the docs for concurrent.futures say: "If a call raises an exception, then that exception will be raised when its value is retrieved from the iterator." 
So switching from multiprocessing might help your #55

etal:
Let's go with concurrent.futures and your current chunking approach, then.
I need to educate myself about this module and then port `batch` to use it,
too.

Is this PR ready to merge, or are there any other changes you have planned
for it before I do so?

brentp:
I think it's ready as long as you understand the travis errors. I didn't want to mess with the .travis.yaml, but it's not getting the futures module for python2.

I didn't have anything that exercised `interval_coverages_count` but `interval_coverages_pileup` seems to be working.

etal:
Great! I added `futures` to the TravisCI conda configuration so the tests ought to pass in the next run.

PR #125, opened 2016-08-30
parallelize segmentation by chromosome.

brentp:
This adds parallelization by chromosome to segment.

I wasn't sure about the save_dataframe with processes>1, for that, can I just do, e.g.
```
s = io.StringIO()
data.to_csv(s)
```

?

### DIFF
--- a/cnvlib/segmentation/__init__.py
+++ b/cnvlib/segmentation/__init__.py
@@ -12,10 +12,41 @@ from .. import core, params, smoothing, tabio, vary
 from ..cnary import CopyNumArray as CNA
 from . import cbs, flasso, haar
 
+from concurrent import futures
+
 from Bio._py3k import StringIO
 
-
 def do_segmentation(cnarr, method, threshold=None, variants=None,
+                    skip_low=False, skip_outliers=10,
+                    save_dataframe=False, rlibpath=None,
+                    processes=1):
+    """Infer copy number segments from the given coverage table."""
+    if processes == 1:
+        return _do_segmentation(cnarr, method, threshold, variants,
+                                skip_low, skip_outliers,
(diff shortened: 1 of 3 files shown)


cnvlib/segmentation/__init__.py
```diff
@@ -12,11 +12,42 @@
...
+                                skip_low, skip_outliers,
+                                save_dataframe, rlibpath)
+
+    chroms = cnarr.chromosome.unique()
+    rets = []
+    # TODO: handle save_dataframe=True
+    with futures.ProcessPoolExecutor(processes) as pool:
+        for ret in pool.map(_ds, ((cnarr[cnarr.chromosome == c], method,
+                                   threshold, variants, skip_low, skip_outliers,
+                                   False, rlibpath) for c in chroms)):
```
etal: You can do:

    ((chromarr, method, ...) for _c, chromarr in cnarr.by_chromosomes())

and avoid creating `chroms` above. (Not that this is a bottleneck, but it's cleaner.)

cnvlib/segmentation/__init__.py
```diff
@@ -12,11 +12,42 @@
...
+            rets.append(ret)
```
etal: Does this work differently than e.g. `rets = list(pool.map(...))`?

etal:
For `save_dataframe`: That option is to print out the CBS dataframe as it would look in R, with minimal contact in Python -- mainly for debugging and exploration. So if you collect the `seg_out` strings from each chunk and then concatenate them however you can, that's ideal.

brentp:
thanks for the review and comments. Latest commit addresses all AFAICT.

etal:
Looks great, thanks so much!

PR #128, opened 2016-09-08
Avoid numpy error with zero weights

chapmanb:
Works around edge case identified during testing latest development
where numpy.average errors out if all the weights sum to zero. This
happened on a not-so-good sample and this workaround avoids failing
outright.


etal:
Thanks!

PR #178, opened 2017-01-10
Update commands.py

myronpeto:
Edited the export_theta args so that it's attribute would be "--min-variant-depth" as opposed to "--min-depth". This makes it congruent with the other export functions and gets rid of the error when trying to export to the theta format.


etal:
Thanks!

PR #239, opened 2017-08-03
diagram: added --title option

micknudsen:
Would be nice to add custom title to diagram plot just as for scatter plot.


etal:
Thanks!

Issue #254, opened 2017-09-15
conda build errors on travis


micknudsen:
There seems to be some problems installing _cnvkit_ using _conda_. It just hangs indefinitely.

```
[micknudsen@work:~]$ conda create -n test cnvkit
Fetching package metadata ...............
Solving package specifications: .
```

Removing **conda-forge** from the list of channels allows me to install _cnvkit_ (but it then fails at runtime). Any clue what could be wrong?

etal:
Sorry, I'm not sure what's causing it to hang for so long. I saw your pull request to bioconda: bioconda/bioconda-recipes#5926

I've noticed that sometimes this command takes a ludicrously long time, but does eventually complete. Maybe up to 90 minutes, going by the TravisCI build logs in the linked issue. Is there a way to make conda emit more debugging info when resolving package dependencies, so I can see which dependency is causing problems?

As a workaround, specifying `cnvkit=0.8.5` ought to retrieve the last known-good conda build. As another workaround, you can install CNVkit's dependencies in a conda environment, then use `pip install -e .` in a clone of the GitHub repo to install the development version within the same conda environment.

When you drop the conda-forge channel, which version of CNVkit do you get?

micknudsen:
I will try and let it run and see what happens. Didn't expect it to take *that* long. Installing without **conda-forge**

`conda create --override-channels -c defaults -c bioconda -c r -n test cnvkit`

installs `0.8.6a0-py36_2`, but then

```
(test) [michaelk@fe1 ~]$ cnvkit.py version
Traceback (most recent call last):
  File "/home/michaelk/anaconda/envs/test/bin/cnvkit.py", line 8, in <module>
    from cnvlib import commands
  File "/home/michaelk/anaconda/envs/test/lib/python3.6/site-packages/cnvlib/__init__.py", line 2, in <module>
    from .cmdutil import read_cna as read
  File "/home/michaelk/anaconda/envs/test/lib/python3.6/site-packages/cnvlib/cmdutil.py", line 7, in <module>
    from skgenome import tabio
  File "/home/michaelk/anaconda/envs/test/lib/python3.6/site-packages/skgenome/__init__.py", line 1, in <module>
    from . import tabio
  File "/home/michaelk/anaconda/envs/test/lib/python3.6/site-packages/skgenome/tabio/__init__.py", line 17, in <module>
    from . import bedio, genepred, gff, picard, seg, tab, textcoord, vcfio
[...]

PR #262, opened 2017-10-16
reference: improve gender checks when chrX missing

chapmanb:
This fixes gender checking based on small test datasets with no chrX
data:

- Always infer sexes, even if female_samples set (otherwise `sexes`
  dictionary not available downstream)
- Map female samples as true based on specification.
- Skip `guess_xx` if missing regions to evaluate

### DIFF
--- a/cnvlib/reference.py
+++ b/cnvlib/reference.py
@@ -25,14 +25,13 @@ def do_reference(target_fnames, antitarget_fnames=None, fa_fname=None,
         logging.info("No FASTA reference genome provided; "
                      "skipping GC, RM calculations")
 
-    if female_samples is None:
-        # NB: Antitargets are usually preferred for inferring sex, but might be
-        # empty files, in which case no inference can be done. Since targets are
-        # guaranteed to exist, infer from those first, then replace those
-        # values where antitargets are suitable.
-        sexes = infer_sexes(target_fnames, male_reference)
-        if antitarget_fnames:
-            sexes.update(infer_sexes(antitarget_fnames, male_reference))
+    # NB: Antitargets are usually preferred for inferring sex, but might be
+    # empty files, in which case no inference can be done. Since targets are
+    # guaranteed to exist, infer from those first, then replace those
+    # values where antitargets are suitable.
+    sexes = infer_sexes(target_fnames, male_reference, female_samples)
+    if antitarget_fnames:
+        sexes.update(infer_sexes(antitarget_fnames, male_reference, female_samples))
 
     # Calculate & save probe centers
     ref_probes = combine_probes(target_fnames, fa_fname,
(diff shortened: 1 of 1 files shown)


cnvlib/reference.py
```diff
@@ -94,9 +93,12 @@ def infer_sexes(cnn_fnames, is_male_reference):
     for fname in cnn_fnames:
         cnarr = read_cna(fname)
         if cnarr:
-            is_xx = cnarr.guess_xx(is_male_reference)
-            if is_xx is not None:
-                sexes[cnarr.sample_id] = is_xx
+            if female_samples:
```
etal: Here, `female_samples` can be True (samples are known XX), False (samples are known XY), or None (not specified by user). Untested, but I think this block should be:

    if female_samples is not None:
        sexes[cnarr.sample_id] = female_samples

chapmanb: Thanks so much for reviewing this. I don't totally understand how `female_samples` and `male_reference` are meant to interact but my intuition was that `female_samples` being true meant that all samples were female. So the behavior I tried to encode here was to make everything female if true, otherwise to look up on a case by case bases using `is_male_reference` and the chromosome inference. Are all samples male if `female_samples` is False? If so, I can swap over the logic. Thanks again for the help.

etal: Yes, both of those are correct. This comes from the command line option `-x (Female|Male)`, where the user can specify that all samples are female (True) or male (False). If it's not used at all, it comes in with the value None, and in that case we guess each sample's chromosomal sex individually. The option is there in case inference is wrong and/or the user knows the sample sexes, e.g. they're studying prostate cancer.

The `is_male_reference` value is just True or False, and it determines whether the reference should have an X ploidy of 1 or 2. If 1 (command line flag `-y`), then the X coverage of "female" input samples is halved, otherwise the X coverage of "male" samples is doubled.

In the prostate cancer, reasonable options would be `-x Male -y` to specify that all input samples are male (`-x Male`), and that the output reference should have an X ploidy of 1 (`-y`). For ovarian cancer you'd use `-x Female` by itself, and if you're used to array CGH, you'd use `-y` by itself and let inference happen. Generally, users ignore these options until they need them. The same two options appear in several other commands and have the same effect, but the terminology is most confusing in `reference`.

Anyway, I thought I'd written this so that if `female_samples` is specified (True or False, not None), the initial pass of loading .cnn files to infer sex is skipped, and later if the `sexes` dictionary then empty, a sample's sex is the value of `female_samples` -- maybe I missed a `sexes.get(sample_id, female_samples)` somewhere.

etal:
I've pushed a commit that I think addresses at least one of the original issues.

In `guess_xx`, if X and Y are both missing then it logs a warning and continues ahead in a fairly reasonable way, without doing much unnecessary work. Are you looking to suppress that logged message?

etal:
And, could you post or describe a test case for this?

(Don't worry about the error on TravisCI, it's currently out of commission due to a bioconda build issue: #254 )

chapmanb:
Eric -- thanks for all the details and the fix. That works great for my failing case. I'm not sure this is worth replicating, it's only a small test case in bcbio used for unit tests so not likely reflective of anything real but it's useful to have in for testing that all the mechanisms are in place. Thanks again for the discussion and quick resolution.

PR #269, opened 2017-10-23
Allow specifying sample_id to 'fix' command

chapmanb:
Eric;
This is a small addition to explicitly pass the sample ID to the fix command, since otherwise it requires that the files are named like sample.targetcoverage.cnn and sample.antitargetcoverage.cnn. I'm trying to generalize our CNVkit usage in bcbio to enable other callers to take advantage of the segmentation and depth so the files end up named slightly differently.


etal:
Makes sense to me. Thanks!

PR #273, opened 2017-11-01
pandas 0.21.0 support: numpy scalar => int

chapmanb:
The most recent pandas release changed how indices are returned, from
numpy int64 to plain python ints:

https://pandas.pydata.org/pandas-docs/version/0.21/whatsnew.html#iteration-of-series-index-will-now-return-python-scalars

In a couple of places cnvkit assumed numpy values and converted them back
into standard ints using `.item`:
```
  for subseq in fasta_extract_regions(fa_fname, cnarr)]
File "/usr/local/share/bcbio-nextgen/anaconda/lib/python2.7/site-packages/cnvlib/reference.py", line 326, in fasta_extract_regions
    yield fa_file[_chrom][start.item():end.item()]
AttributeError: 'int' object has no attribute 'item'
```
This swaps over to use use `int` conversion which should work on both
older and newer pandas.

### DIFF
--- a/cnvlib/reference.py
+++ b/cnvlib/reference.py
@@ -326,7 +326,7 @@ def fasta_extract_regions(fa_fname, intervals):
         for chrom, subarr in intervals.by_chromosome():
             logging.info("Extracting sequences from chromosome %s", chrom)
             for _chrom, start, end in subarr.coords():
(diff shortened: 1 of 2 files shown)


etal:
Thanks!

PR #285, opened 2017-11-21
call cn filter: handle blocks with zero weights

chapmanb:
When running `cnvkit.py call --filter cn`, some edge cases have
blocks with zero weights, causing errors normalizing with these:

https://github.com/chapmanb/bcbio-nextgen/issues/2112#issuecomment-345516176

This falls back to using unweighted averages and means when weights
aren't usable.

### DIFF
--- a/cnvlib/segfilters.py
+++ b/cnvlib/segfilters.py
@@ -72,18 +72,33 @@ def squash_region(cnarr):
            'start': cnarr['start'].iat[0],
            'end': cnarr['end'].iat[-1],
           }
-    out['log2'] = np.average(cnarr['log2'], weights=cnarr['weight'])
+    if cnarr['weight'].sum() > 0:
+        out['log2'] = np.average(cnarr['log2'], weights=cnarr['weight'])
+    else:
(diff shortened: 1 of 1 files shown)


etal:
Thanks! Any idea how the 0-weight segments were created?

chapmanb:
Thanks, I'm not sure how it got generated but sent the file off list in case that gives you any clues. Much appreciated.

etal:
It looks like these zero-weight segments came from underlying bins with zero coverage depth. These were then segmented out separately -- one with the segment covering just 1 bin, the other covering 4 bins, out of 2380 segments, so this is a pretty rare occurrence that is unlikely to be seen in tumor samples when `--drop-low-coverage` is used. With a paired or pooled reference these bins would have been filtered out upstream, but not with a flat (control-free) reference. Maybe that's what happened here. I'll see about handling this situation better during segmentation.

Thanks again for the patch.

chapmanb:
Brilliant, thanks for identifying the underlying issue and helping work on it. Hopefully with the new improvements in bcbio integration we'll see less of this. I appreciate you helping track it all down.

PR #294, opened 2017-12-20
Fix bug in README.rst

MajoroMask:
Little bug in the README.


MajoroMask:
R use c() function to put multiple strings into one object, then feed it to another function as one argument.
`biocLite("DNAcopy", "cghFLasso")` actually means `biocLite(pkgs = "DNAcopy", suppressUpdates = "cghFLasso")`

etal:
Good catch, thanks!

PR #391, opened 2018-10-21
vcfio: support GATK inputs with empty ALT alleles

chapmanb:
Recent versions of GATK 4 can output lines with empty ALT alleles,
thus missing an ALT AD value. This checks for this case, correctly
returning zero and adds an example to the tests.

### DIFF
--- a/skgenome/tabio/vcfio.py
+++ b/skgenome/tabio/vcfio.py
@@ -268,7 +268,13 @@ def _get_alt_count(sample):
     if sample.get('AD') not in (None, (None,)):
         # GATK and other callers: (ref depth, alt depth)
         if isinstance(sample['AD'], tuple):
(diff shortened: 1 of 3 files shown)


etal:
Cool cool. What's the origin and interpretation of these VCF lines? In the test file it says the total allele count is 2, and the alt allele could maybe be interpreted as "unknown non-reference". But the coverage depth and quality there looks fine, so I'm wondering what edge case GATK4 is attempting to handle here.

chapmanb:
Eric;
These don't convey much useful information. My guess it that it's a temporary version specific bug, so I mostly just didn't want CNVkit to be failing on them. I hadn't investigated further but probably will ask the Broad team if we continue to see them in future versions. Thanks so much for merging the fix.

Issue #397, opened 2018-11-14
error while performing test


cnizo:
Hi @etal!

I ran into an error while performing the test. Here is the log:
```
Treating sample p2-5_5.call as male
Treating sample p2-9_2.call as male
Treating sample p2-20_1.call as male
Treating sample p2-20_2.call as male
Treating sample p2-20_3.call as male
Treating sample p2-20_4.call as male
Treating sample p2-20_5.call as male
Wrote p2-all.bed
python ../cnvkit.py export vcf -o p2-9_2.vcf -y --cnr build/p2-9_2.cnr build/p2-9_2.call.cns -y -o p2-9_2.vcf
Treating sample p2-9_2.call as male
Traceback (most recent call last):
  File "../cnvkit.py", line 13, in <module>
    args.func(args)
  File "/home/varianttools/bioinfo/app/cnvkit/cnvlib/commands.py", line 1625, in _cmd_export_vcf
    is_sample_female, args.sample_id, cnarr)
  File "/home/varianttools/bioinfo/app/cnvkit/cnvlib/export.py", line 237, in export_vcf
    segments = assign_ci_start_end(segments, cnarr)
  File "/home/varianttools/bioinfo/app/cnvkit/cnvlib/export.py", line 261, in assign_ci_start_end
    ci_lefts, ci_rights = zip(*lefts_rights)
  File "/home/varianttools/bioinfo/app/cnvkit/cnvlib/export.py", line 260, in <genexpr>
    for _seg, bins in cnarr.by_ranges(segarr, mode="outer"))
  File "/home/varianttools/_env/lib/python3.5/site-packages/pandas/core/indexing.py", line 1869, in __getitem__
    return self.obj._get_value(*key, takeable=self._takeable)
  File "/home/varianttools/_env/lib/python3.5/site-packages/pandas/core/series.py", line 928, in _get_value
    return _maybe_box_datetimelike(self._values[label])
IndexError: index 0 is out of bounds for axis 0 with size 0
Makefile:148: recipe for target 'p2-9_2.vcf' failed
make: *** [p2-9_2.vcf] Error 1
```
Do you have an idea of what it might be? I have just cloned the repository and I have already tried doing``` make clean; make all```

etal:
Thanks for letting me know. It looks like `export vcf` thinks one of its input files, or some component, is empty. Did segmentation work correctly, i.e. the .cns files list multiple segments, at least one per chromosome?

cnizo:
[...]

PR #408, opened 2018-12-11
do_scatter now returns a figure

jeremy9959:
Because the scatter command sometimes returns a compound figure, I've set it up so that do_scatter returns a figure object rather than an axis object.  

I've only partially tested this.  I don't have any snv data to try it with so I wasn't able to test the situation where the plot contains both snv and cnv data.

### DIFF
--- a/cnvlib/scatter.py
+++ b/cnvlib/scatter.py
@@ -39,19 +39,19 @@ def do_scatter(cnarr, segments=None, variants=None,
         MB = 1
 
     if not show_gene and not show_range:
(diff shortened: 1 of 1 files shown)


jeremy9959:
I see your CI checks failed, but looking at the log, the issues don't appear to have anything to do with scatter.py, and that's the only file I touched.

etal:
Yes, this looks good to me, and I think the unit test failures are a different issue (#397). Thanks for this contribution!

jeremy9959:
Thanks for developing cnvkit, it's a very impressive piece of work!

PR #454, opened 2019-06-10
Adding --smooth-cbs parameter for cnvkit.py segment functionality

EwaMarek:
Hi,

This pull request adds --smooth-cbs parameter to cnvkit.py segment functionality. It performs smoothing of the data using DNAcopy::smooth.CNA before segmenting using CBS method. 

Without this parameter the smoothing function is used only in order to detect and drop outliers, but the data on which the segmentation is performed is not smoothed. Adding this parameter seems to increase the sensitivity for CNV calling in targeted sequencing and thus may be useful for some people.

Please let me know what do you think about it.

Best wishes,
Ewa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,44 +1,47 @@
-language: c
-sudo: false
+language: python
 
-os:
-    - osx
-    - linux
-env:
-    # Earliest supported versions
-    - CONDA_PY=3.5 PANDAS=0.22.0 PYSAM=0.10.0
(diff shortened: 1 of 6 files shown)


etal:
Thanks for this contribution, and sorry I missed it earlier.

The HMM method also uses a smoothing step before segmentation. Maybe these could be consolidated. I'll merge this PR and then see about it.

PR #493, opened 2020-01-23
Make Docker file functional now only supports py3

keiranmraine:
Changes apt installs from `python-*` to `python3-*` also:

* add Cyton install
* ensure doesn't run as root (for singularity compatibility)


etal:
Thanks for this update!

PR #495, opened 2020-01-31
Cram support

sridhar0605:
Cram support for CNVkit 0.9.7.b1
Dependencies version:
```
pysam==0.15.3
biopython==1.75
matplotlib==3.0.3
numpy==1.17.3
pandas==0.25.3
pyfaidx==0.5.5.2
pysam==0.15.3
reportlab==3.5.34
scipy==1.4.1
pomegranate==0.12.0
```

Tested this on HCC1395 data set(both bams and crams) hosted here  [https://pmbio.org/module-02-inputs/0002/05/01/Data/](HCC1395) Aligning to B38. 

Only warning i see is 
`[W::find_file_url] Failed to open reference "https://www.ebi.ac.uk/ena/cram/md5/6aef897c3d6ff0c78aff06ac189178dd": Protocol not supported` which is referenced [https://github.com/pysam-developers/pysam/blob/master/tests/AlignmentFile_test.py#L2415](here) as an error but its not really an error and does not break any thing and may be result of how pysam was installed on my machine.

Thanks
Sid

### DIFF
--- a/cnvlib/coverage.py
+++ b/cnvlib/coverage.py
@@ -157,6 +157,7 @@ def interval_coverages_pileup(bed_fname, bam_fname, min_mapq, procs=1):
     logging.info("Processing reads in %s", os.path.basename(bam_fname))
     if procs == 1:
         table = bedcov(bed_fname, bam_fname, min_mapq)
+        print(bed_fname, bam_fname, min_mapq)
     else:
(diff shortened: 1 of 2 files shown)


etal:
This looks great to me. Thanks so much!

PR #536, opened 2020-08-28
Fix a bug when plotting scatters with region specified

### LINKED ISSUE #457: scatter finds zero probes when using regions
I have a problem with version 0.9.6. I can plot single chromosomes using the following command:

`cnvkit.py scatter -s Sample.cn{s,r} -c 17 -o 17.png` <- this works

But if I want to plot I defined area say

`cnvkit.py scatter -s Sample.cn{s,r} -c 17:37000000-38000000 -o 17.png` <- this doesn't

Showing 0 probes and 1 selected genes in region 17:35999999-39000000

Similarily if I try to plot only a selected gene.

`cnvkit.py scatter -s Sample.cn{s,r} -g ERBB2`

Showing 0 probes and 1 selected genes....

I tried with an older installation, version 0.9.1 and all the above commands works just fine

tskir:
Closes #457.

This resolves a plotting bug in skgenome/intersect.py, which concerns two functions:
* `iter_ranges` is supposed to take the dataframe and return the **bins** which come from a certain region (defined with chrom+starts+ends).
* `idx_ranges` is supposed to also take the dataframe and return the **indices** of those same bins.

`iter_ranges` passes the dataframe to `idx_ranges`, accepts the incides, fetches the bins using them, and returns the bins.

Both `iter_ranges` and `idx_ranges` can currently work either with _chrom_ specified, or without one. The bug appears when `iter_ranges` is used with _chrom_ specified. It passes the parameters to `idx_ranges` and expects the indices for the **original** dataframe. However, `idx_ranges` in this case simply applies filtering by _chrom_  (`table = table[table['chromosome'] == chrom]`) and returns the indices from that **filtered** dataframe. `iter_ranges`, once it receives the incides, tries to apply them to the original dataframe, causing the problem.

The simplest solution which I found is to simply move filtering by _chrom_ from `idx_ranges` to `iter_ranges`, and get rid of the _chrom_ parameter in `idx_ranges` entirely. Note that currently, `idx_ranges` is only called twice in the whole code:
1. From `iter_ranges` (with _chrom_ specified);
1. From some other function, where _chrom_ is always None.

This means that by removing filtering by _chrom_ from `idx_ranges`, we definitely aren't breaking anything else in the code, but are fixing the `iter_ranges`: now it accepts the full dataframe, applies filtering, and then uses `idx_ranges` to apply _filtered_ indices to a _filtered_ dataframe.

I have additionally verified that this modification does indeed fix the plotting bug described in #457.

### DIFF
--- a/skgenome/intersect.py
+++ b/skgenome/intersect.py
@@ -76,7 +76,14 @@ def into_ranges(source, dest, src_col, default, summary_func):
 def iter_ranges(table, chrom, starts, ends, mode):
     """Iterate through sub-ranges."""
     assert mode in ('inner', 'outer', 'trim')
-    for region_idx, start_val, end_val in idx_ranges(table, chrom, starts, ends,
+    # Optional if we've already subsetted by chromosome (not checked!)
+    if chrom:
+        assert isinstance(chrom, str)  # ENH: accept array?
+        try:
+            table = table[table['chromosome'] == chrom]
+        except KeyError:
+            raise KeyError("Chromosome %s is not in this probe set" % chrom)
+    for region_idx, start_val, end_val in idx_ranges(table, starts, ends,
             'inner' if mode == 'inner' else 'outer'):
         subtable = table.iloc[region_idx]
         if mode == 'trim':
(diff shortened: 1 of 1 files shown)


tskir:
P.S. Shout out to @ViktorHy, who originally reported the issue, and to @tetedange13, who performed part of the investigation, based on which I could build my solution

etal:
Thanks very much for this fix! It will be in the next CNVkit release.

Issue #547, opened 2020-10-21
Bin log2 ratios missing due to preceding gc-masked bins


tridgley:
Summary: ~1% of the bins in our sample.cnr files do not have log2 ratios
Version: CNVkit 0.9.7

Details: The CNVkit batch pipeline was initially run for creating a flat reference based on the hg38 genome and calling CNV on aligned WGS data for tumor samples only (no normal samples):
/tridgley/wgs_sequences/CNVcalls/$ python3 cnvkit.py batch sample.bam --normal --fasta ../reference/hg38/genome.fa --output-reference ../CNVref/hg38_flat_reference.cnn --method wgs

For CNV calling of additional samples thereafter, we used the hg38_flat_reference:
/tridgley/wgs_sequences/CNVcalls/$ python3 cnvkit.py batch sample.bam -r ../CNVref/hg38_flat_reference.cnn -p 0

At the fix step of the batch pipeline, 4862 bins were thrown out due gc>0.7 or gc<0.3 according to mask_bad_bins in fix.py:
`if 'gc' in cnarr:`
`    mask |= (cnarr['gc'] > .7) | (cnarr['gc'] < .3)`
`    return mask`
This line was output, as expected:
`Keeping 582687 of 587549 bins`

Let bb_i be the position of each bad bin identified by the fix. After the masking step, the next good bin at position bb_i+i gets an empty log2 ratio in the cnr file. In our specific case, there were 4724 bins with missing log2 ratios. For instance, here is the first bad bin in our genome that leads to the problem being reported:

hg38_flat_reference.cnn (Bad bin at 996335-1001339 has gc>0.7):
`chromosome      start   end     gene    log2    depth   gc      rmask   spread`
`chr1    996335  1001339 -       0       1       0.707834        0.0603517       0`
`chr1    1001339 1006344 -       0       1       0.630569        0.267932        0`
`chr1    1006344 1011348 -       0       1       0.519185        0.60012 0`

target.cnn (Bins have sufficient coverage and log2 coverage is present):
`chromosome      start   end     gene    depth   log2`
`chr1    996335  1001339 -       18.8086 4.23332`
`chr1    1001339 1006344 -       15.4959 3.95382`
`chr1    1006344 1011348 -       12.5116 3.64519`

sample.cnr (Bad bin at pos bb_1 was masked, but the next good bin 1001339-1006344 at pos bb_1+1 is missing its log2 ratio):
`chromosome      start   end     gene    depth   log2    weight`
[...]

Issue #436, opened 2019-04-15
cnvkit batch fail for WES data 


flower1996:
Dear，
I am running CNVkit to call cnvs on samples sequenced with WES sequencing, and get an error below:
`Segmenting with method 'cbs', significance threshold 0.0001, in 1 processes
Traceback (most recent call last):
  File "/home/miniconda3/bin/cnvkit.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/biosoftware/cnvkit/cnvkit/cnvkit.py", line 9, in <module>
    args.func(args)
  File "/home/biosoftware/cnvkit/cnvkit/cnvlib/commands.py", line 143, in _cmd_batch
    args.cluster)
  File "/home/biosoftware/cnvkit/cnvkit/cnvlib/parallel.py", line 19, in submit
    return SerialFuture(func(*args))
  File "/home/biosoftware/cnvkit/cnvkit/cnvlib/batch.py", line 192, in batch_run_sample
    else {}))
  File "/home/biosoftware/cnvkit/cnvkit/cnvlib/segmentation/__init__.py", line 66, in do_segmentation
    for _, ca in cnarr.by_arm())))
  File "/home/biosoftware/cnvkit/cnvkit/cnvlib/segmentation/__init__.py", line 91, in _ds
    return _do_segmentation(*args)
  File "/home/biosoftware/cnvkit/cnvkit/cnvlib/segmentation/__init__.py", line 162, in _do_segmentation
    seg_out = core.call_quiet(rscript_path, '--vanilla', script_fname)
  File "/home/biosoftware/cnvkit/cnvkit/cnvlib/core.py", line 32, in call_quiet
    % (' '.join(args), err))
RuntimeError: Subprocess command failed:
$ Rscript --vanilla /tmp/tmp7lkqejav

b'Loading probe coverages into a data frame\nWarning message:\nIn CNA(cbind(tbl$log2), tbl$chromosome, tbl$start, data.type = "logratio",  :\n  markers with missing chrom and/or maploc removed\n\nSegmenting the probe data\nError in segment(cna, weights = tbl$weight, alpha = 1e-04) : \n  length of weights should be the same as the number of probes\n\xe5\x81\x9c\xe6\xad\xa2\xe6\x89\xa7\xe8\xa1\x8c\n`.

The command that I run is:
`cnvkit.py batch tumor.bam --normal normal.bam \
	--targets hg38.exon.bed \
	--method amplicon \
[...]

PR #551, opened 2020-11-24
Fix to ensure dataframe indices still match after adjustments

johnegarza:
This PR will hopefully solve, or at least further the conversation on how to solve, the bug reported in #436 and possibly #547. For reference, my command, run within the `etal/cnvkit:0.9.7` docker container, was `/usr/bin/python3 /code/cnvkit.py batch /inputs/1455918947/tumor.bam --normal /inputs/29426665/normal.bam --targets /inputs/876494580/hla_and_brca_genes_bait.interval_list --method hybrid`. I am happy to provide these input files if you'd like. The resulting error trace:

```
Traceback (most recent call last):
  File "/code/cnvkit.py", line 9, in <module>
    args.func(args)
  File "/code/cnvlib/commands.py", line 138, in _cmd_batch
    pool.submit(batch.batch_run_sample,
  File "/code/cnvlib/parallel.py", line 19, in submit
    return SerialFuture(func(*args))
  File "/code/cnvlib/batch.py", line 186, in batch_run_sample
    segments = segmentation.do_segmentation(cnarr, segment_method,
  File "/code/cnvlib/segmentation/__init__.py", line 61, in do_segmentation
    rets = list(pool.map(_ds, ((ca, method, threshold, variants,
  File "/code/cnvlib/segmentation/__init__.py", line 89, in _ds
    return _do_segmentation(*args)
  File "/code/cnvlib/segmentation/__init__.py", line 174, in _do_segmentation
    seg_out = core.call_quiet(rscript_path,
  File "/code/cnvlib/core.py", line 31, in call_quiet
    raise RuntimeError("Subprocess command failed:\n$ %s\n\n%s"
RuntimeError: Subprocess command failed:
$ Rscript --no-restore --no-environ /tmp/tmp6alxd4pm

b'Loading probe coverages into a data frame\nWarning message:\nIn CNA(cbind(tbl$log2), tbl$chromosome, tbl$start, data.type = "logratio",  :\n  markers with missing chrom and/or maploc removed\n\nSegmenting the probe data\nError in segment(cna, weights = tbl$weight, alpha = 1e-04) : \n  length of weights should be the same as the number of probes\nExecution halted\n'
```

The issue begins when [a mask is applied](https://github.com/etal/cnvkit/blob/de22a53dc5360a447637af5c9b6c50fc1f6c32db/cnvlib/fix.py#L74-L75) to drop poor quality bins in `cnvlib/fix.py`. In my case, examination of the dataframes after masking showed that their indices were no longer a continuous series from 0, 1, 2, ... x, but instead had gaps. While there are gaps, at this point in the code, both `cnarr` and `ref_matched` still have matching indices. However, a later [section of the code](https://github.com/etal/cnvkit/blob/de22a53dc5360a447637af5c9b6c50fc1f6c32db/cnvlib/fix.py#L84-L101) may pass `cnarr` to the function `center_by_window` and [set its value to the function's return value.](https://github.com/etal/cnvkit/blob/de22a53dc5360a447637af5c9b6c50fc1f6c32db/cnvlib/fix.py#L93) Within `center_by_window`, [the index of `cnarr` is reset.](https://github.com/etal/cnvkit/blob/de22a53dc5360a447637af5c9b6c50fc1f6c32db/cnvlib/fix.py#L155) No such reset is applied to `ref_matched.` This means that when [the `log2` column of `ref_matched` is subtracted from the `log2` column of `cnarr`,](https://github.com/etal/cnvkit/blob/de22a53dc5360a447637af5c9b6c50fc1f6c32db/cnvlib/fix.py#L52) there are "gaps" in `ref_matched` relative to `cnarr`, and the values are set to `NaN` in `cnarr`. Modified code with output to demonstrate:

Code
```
    logging.info('cnarr log2:')
    logging.info(cnarr.data['log2'])
    logging.info('ref matched log2:')
    logging.info(ref_matched['log2'])
    
    cnarr.data['log2'] -= ref_matched[log2_key]

    logging.info('cnarr, but with gaps:')
    logging.info(cnarr.data['log2'])
```

Logs
```
cnarr log2:
0     0.838088
1    -1.612942
2     0.729092
3     0.092408
...
ref matched log2:
0     0.252140
2     0.000000
3     0.000000
...
cnarr, but with gaps:
0     0.585948
1          NaN
2     0.729092
3     0.092408
```

Thus, `NaN` values are introduced to the `log2` column of `cnarr`, resulting in the Rscript error shown above. This PR adds a flag and conditional that resets the index of `ref_matched` if the index of `cnarr` has been reset by a call to `center_by_window`. This should be safe, since `center_by_window` preserves the original order of the rows, and I believe sufficient to solve the problem; however, there may be some things I missed, and I'd appreciate your thoughts.

### DIFF
--- a/cnvlib/fix.py
+++ b/cnvlib/fix.py
@@ -82,23 +82,29 @@ def load_adjust_coverages(cnarr, ref_cnarr, skip_low,
         logging.warning("WARNING: most bins have no or very low coverage; "
                         "check that the right BED file was used")
     else:
+        cnarr_index_reset = False
         if fix_gc:
             if 'gc' in ref_matched:
                 logging.info("Correcting for GC bias...")
                 cnarr = center_by_window(cnarr, .1, ref_matched['gc'])
+                cnarr_index_reset = True
             else:
                 logging.warning("WARNING: Skipping correction for GC bias")
         if fix_edge:
             logging.info("Correcting for density bias...")
             edge_bias = get_edge_bias(cnarr, params.INSERT_SIZE)
             cnarr = center_by_window(cnarr, .1, edge_bias)
+            cnarr_index_reset = True
         if fix_rmask:
             if 'rmask' in ref_matched:
                 logging.info("Correcting for RepeatMasker bias...")
                 cnarr = center_by_window(cnarr, .1, ref_matched['rmask'])
+                cnarr_index_reset = True
             else:
                 logging.warning("WARNING: Skipping correction for "
                                 "RepeatMasker bias")
+        if cnarr_index_reset:
+            ref_matched.data.reset_index(drop=True, inplace=True)
     return cnarr, ref_matched
 
 


etal:
Thanks so much for triaging this issue and submitting a fix! Your description makes sense and the code looks good to me, so I've merged it.

PR #555, opened 2020-12-03
Cram reference support

johnegarza:
This PR adds support for explicitly specifying a reference fasta file to use when opening cram files. Previously, cnvkit defaulted to attempting to look up the reference sequences via md5 hash online, which pysam does not support. The first commit in this PR was part of my test workflow (the same as in #551) and has been tested extensively. The rest of the changes were made by finding calls to pysam that involved opening cram files and working backwards; I have tested these as best as I could, but may have missed some code paths, and would appreciate your input. I have verified that my modified calls to pysam with no fasta (`reference_filename=None`) function normally.


etal:
Thank you for this nicely formatted change set. The code here looks good to me; I'll merge it.

Issue #579, opened 2021-03-16
Why some of my diagram figs have extra genes in chr1?


HYan-lei:
Hi, 

I run all the samples together. Most samples's diagram figs are normal, but few of them have extra gene names in chr1. I used the cnvkit  cnvkit 0.9.7.b1. Can you please help me with this issue? Thanks. HY
[19_BLE_65.diagram.pdf](https://github.com/etal/cnvkit/files/6152842/19_BLE_65.diagram.pdf)

tskir:
Hi @HYan-lei,

A number of plotting bugs were recently fixed in CVNKit version 0.9.8, and this one looks like it should be one of them. Could you please upgrade to the latest version and see if this resolves the problem?

HYan-lei:
Thanks. I just tried the 0.9.8 version, and still has the same issue.

tskir:
@HYan-lei You're right! It's a different one, I was able to reproduce it.

It turned out to have the same underlying reason as issues #573 and #576.

As a result, it will be fixed by the PR I just submitted, #580.

Thank you for reporting it!

tskir:
Oh, and actually, the reason you're only seeing extra genes in chr1 is because the indexing function was broken for all chromosomes **except** 1. So with the fix, you will very likely see lots of extra genes for all chromosomes.

The actual reason you're seeing lots of extra genes is probably the samples themselves. In my experience, for every batch of samples (depending on technology, protocols etc.), there will be a small number of "noisy" samples which appear to have lots and lots of CNVs. This usually just means that the coverage profile for them is different from the rest of the batch.

This issue will close when the PR is merged, but if you continue to have problems, please do not hesitate to open a new one.

HYan-lei:
What you mean "when the PR is merged?" Thanks.

tskir:
This means I've submitted a code change which will fix this issue (#580), and it's now up to the author of CNVkit to review and merge it. Once this is done and the new version is released, the issue will be resolved.

HYan-lei:
great. Thanks. I will try to run again when the new version is released. Thanks for the help.

PR #580, opened 2021-03-18
Fix subgary indexing in cnary.by_gene()

### LINKED ISSUE #573: genemetrics maybe have bugs in 0.9.7
Hi,
 when I use genemetrics with cnvkit 0.9.7 , the variants are disappeared, which can be detected with cnvkit 0.9.6.

Then I check the scripts, the differents as follows :
 a. The function "as_dataframe" in  skgenome/gary.py is different between 0.9.7 and 0.9.6
 b. In  0.9.6 the dataframe index is reset，which is not reset by default in 0.9.7
 c. So the script of function "by_gene" in cnary.py "yield gene, subgary.as_dataframe(subgary.data.iloc[start_idx:end_idx])" return the none

### LINKED ISSUE #576: genemetrics completing without error part way through chromosome one
I am using a conda installed cnvkit 0.9.8 environment (and this has also been tested on a conda installed cnvkit 0.9.7 environment, with the same result).

_cnvkit.pl batch_ runs successfully, producing a genome-wide .cnr and .cns per sample.

_cnvkit.pl genemetrics_ ran with e.g. ..

`cnvkit.py genemetrics $cnr -s $cns -t 0.1 -m 3 --output $output_fn`

.. also claims to run successfully, printing the output:

```
Treating sample [ $ID ] as female
Found 108 gene-level gains and losses
Wrote [ $output_fn ]
```

However the 108 lines in $output_fn only cover up to chr1 16592382.

E.g. tail $output_fn

```
SRM	chr1	15944260	15947322	-0.254386	349.365	5.9755	-0.250492	-0.259758	7	1660.24	2111
EXOSC10	chr1	15948134	16029878	-0.254386	365.13	20.4798	-0.250492	-0.259758	27	1660.24	2111
MTOR	chr1	16029943	16217204	-0.254386	391.329	55.5135	-0.250492	-0.259758	73	1660.24	2111
UBIAD1	chr1	16229477	16250183	-0.254386	339.729	5.65058	-0.250492	-0.259758	7	1660.24	2111
DISP3	chr1	16250969	16404047	-0.254386	313.915	23.4543	-0.250492	-0.259758	29	1660.24	2111
FBXO2	chr1	16408865	16410975	-0.254386	268.668	2.77106	-0.250492	-0.259758	5	1660.24	2111
FBXO44	chr1	16411475	16428079	-0.254386	418.085	4.39503	-0.250492	-0.259758	6	1660.24	2111
FBXO6	chr1	16431081	16450042	-0.254386	565.949	5.6832	-0.250492	-0.259758	7	1660.24	2111
MAD2L2	chr1	16450047	16535372	-0.254386	255.099	6.02192	-0.250492	-0.259758	8	1660.24	2111
DRAXIN	chr1	16536708	16592382	-0.254386	506.403	7.49921	-0.250492	-0.259758	12	1660.24	2111
```

The same thing occurs for all samples (up to varying chr1 positions).

Can you shed any light on the problem here? The results are the same regardless of the number of cores provided to the task.

tskir:
Closes #573. Closes #576. Thank you @diushiguzhi @eriktoo @hrkemp @drmrgd for reporting and initial investigations!

cnary.by_gene() [iterates over chromosomes,](https://github.com/etal/cnvkit/blob/bb519a45eaa79fb7b0c904d83bdb2c1b85607c1f/cnvlib/cnary.py#L87) which are represented as dataframe slices which [preserve the original indexes](https://github.com/etal/cnvkit/blob/bb519a45eaa79fb7b0c904d83bdb2c1b85607c1f/skgenome/gary.py#L259). Because of this, accessing them by `.iloc` is not going to be possible, and `.loc` needs to be used instead. (An alternative solution would be to keep `.iloc` but reset the sub-dataframe index somewhere along the way, but I don't see a good reason to do this.)

The autotests don't seem to be working but I've ran the full suite locally and everything's fine.

### DIFF
--- a/cnvlib/cnary.py
+++ b/cnvlib/cnary.py
@@ -83,7 +83,6 @@ class CopyNumArray(GenomicArray):
             Pairs of: (gene name, CNA of rows with same name)
         """
         ignore += params.ANTITARGET_ALIASES
-        start_idx = end_idx = None
         for _chrom, subgary in self.by_chromosome():
             prev_idx = 0
             for gene, gene_idx in subgary._get_gene_map().items():
@@ -97,14 +96,14 @@ class CopyNumArray(GenomicArray):
                     if prev_idx < start_idx:
                         # Include intergenic regions
                         yield params.ANTITARGET_NAME, subgary.as_dataframe(
-                                subgary.data.iloc[prev_idx:start_idx])
+                                subgary.data.loc[prev_idx:start_idx])
                     yield gene, subgary.as_dataframe(
-                            subgary.data.iloc[start_idx:end_idx])
+                            subgary.data.loc[start_idx:end_idx])
                     prev_idx = end_idx
             if prev_idx < len(subgary) - 1:
                 # Include the telomere
                 yield params.ANTITARGET_NAME, subgary.as_dataframe(
-                        subgary.data.iloc[prev_idx:])
+                        subgary.data.loc[prev_idx:])
 
     # Manipulation
 


cnvlib/cnary.py
```diff
@@ -83,7 +83,6 @@ def by_gene(self, ignore=params.IGNORE_GENE_NAMES):
             Pairs of: (gene name, CNA of rows with same name)
         """
         ignore += params.ANTITARGET_ALIASES
-        start_idx = end_idx = None
```
tskir, line 86: Removed this line because it's never used—both `start_idx` and `end_idx` are initialised again below

tskir:
P. S. It also solves the problem with the chromosome diagrams in #579 because the underlying reason is exactly the same.

etal:
Do you see any other related code where `iloc` or `loc` has problems on original vs. reset index differences? I remember addressing something similar in the modules `skgenome/intersect.py` and `skgenome/merge.py`.

The fixes in `by_gene()` look correct, so I'll merge this PR and keep an eye out for any similar issues elsewhere in the codebase that haven't already been addressed.

PR #586, opened 2021-03-24
Rename `n_bins` to `probes` in genemetrics output

### LINKED ISSUE #585: cnvkit export vcf does not work when nbins is used instead of probes in cns header
cnvkit version: 0.9.8

In my pipeline, after generating .cnr files (diploid X) they are processed by genemetrics without segments, then segmetrics (with --gender female to avoid shifting X log2s) to get statistics, and finally with call. Exporting the final call .cns to bed format works fine, but when exporting to vcf the files contain only the header and no cnvs. When running a pipeline that does normal segmentation instead of genemetrics export to vcf works. The only difference in the header is that the genemetrics case has a column 'n_bins' and the regular segmentation a column 'probes'.

The responsible line in export.py is 266, which does not check for the presence of n_bins or probes and assumes the latter:

`    out_dframe = segments.data.reindex(columns=["chromosome", "end", "log2", "probes"])`

A quick fix is to run the files with n_bins through sed to convert to probes:

`sed s/n_bins/probes/ $file`

tskir:
Closes #585. Thank you @eriktoo for reporting and investigating this.

The column `n_bins` in the genemetrics output means exactly the same thing as `probes` in regular CNS files. For compatibility, it is better if they are named the same. This would also allow feeding genemetrics CNS files to the call and export commands.

### DIFF
--- a/cnvlib/diagram.py
+++ b/cnvlib/diagram.py
@@ -101,7 +101,7 @@ def _get_gene_labels(cnarr, segarr, cnarr_is_seg, threshold, min_probes):
     else:
         # Only bin-level ratios (.cnr)
         rows = reports.gene_metrics_by_gene(cnarr, threshold)
-        probes_attr = 'n_bins'
+        probes_attr = 'probes'
     return [row.gene for row in rows if getattr(row, probes_attr) >= min_probes]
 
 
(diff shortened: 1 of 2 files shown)


tskir:
Note: in theory, `n_bins` is a name which better reflects reality than `probes`. However, it's much easier to change it in this direction, since renaming `probes` to `n_bins` would involve quite a lot of changes throughout the pipeline which I'd rather not do at this stage.

etal:
This looks like a good solution, thanks!

PR #590, opened 2021-04-05
Fix smoothing bug when signal is shorter than a window

### LINKED ISSUE #587: [CNVkit v0.9.8] Errored smoothing with HAAR-segmentation ?
Hi,
<br>
First, **thanks** for this very useful tool and your investment developping it !

*Informatic context:* Running CNVkit **v0.9.8** on CentOS 6 virtual machine, installed through:
1) `conda create -n my_env pip`
2) `conda activate my_env && pip install cnvkit` (+ conda installation of 'r-base' for DNAcopy R package)
<br>

*Data:* 
Hybrid **capture panel of genes and hotspots** (capture probes coordinates used as manifest), somatic DNA (FFPE) with **no normal sample**
I was previously using CNVkit v0.9.6 on the same panel but with an amplicon tech. Segmenting with HAAR used to give better results than CBS, so I wanted to give it a try on my new data and with an upgraded CNVkit.
=> I should precise that using v0.9.8 on my hybrid-capture data with a **CBS-segmentation went fine** (so no installation issues I guess ?)
<br>

*Exact command and error:*
`cnvkit.py batch my_patient.bam -p 1 -n --segment-method haar -f hg19.fa -t capture_probes.manifest.bed --drop-low-coverage`
Most patient when fine, except for a couple, that raised following error:
```
  File "/home/bioinfo/miniconda3/envs/cnvkit_repo/lib/python3.9/site-packages/cnvlib/segmentation/haar.py", line 328, in HaarConv
      highNonNormed += signal[highEnd] * weight[highEnd] - signal[k-1] * weight[k-1]
IndexError: index 2 is out of bounds for axis 0 with size 2
```
(put [complete traceback here](https://github.com/etal/cnvkit/files/6197925/tracebak.txt) for clarity)
<br>

*Possible cause:*
This error happens during **segmentation of chrY** and can be reproduced: `cnvkit.py segment --drop-low-coverage -m haar my_patient.cnr` (I think I can even share with you both a buggy and a fonctionning ".cnr" with equivalent chrY depth, if needed)
My panel has a single probe on SRY used as a control, which is "binned" by CNVkit into 1 antitarget and 1 target region within each .cnr as examplified here:
```
Y  2507140  2654431  Antitarget                               0        -10.8612  0.0001
Y  2654931  2655203  NM_003140|SRY|3PUTR|1;NM_003140|SRY|1/1  1286.11  0.110645  0.932772
```
<br>

- Every patient "HAAR-segmentable" has a **0 mean_depth for SRY antitarget region** and a "classic" depth for target region (> 1000X for male and around 0 for female)
=> low-covered antitarget is excluded lead to a 1-sized signal_array, so smoothing returns a `np.zeros(signalSize, dtype=np.float_)` thanks to this condition
https://github.com/etal/cnvkit/blob/b218280edb266788ea8326596d4283183ce425ea/cnvlib/segmentation/haar.py#L289
- In the contrary, my bogued patients have a very small depth at their antitarget SRY region (0.000536353, but not 0)
=> So antitarget is kept and this leads to a 2-sized signal_arr, next given to `smoothing.savgol()` *via* [cnarr.smooth_log()](https://github.com/etal/cnvkit/blob/b218280edb266788ea8326596d4283183ce425ea/cnvlib/cnary.py#L476) and after "winging", we obtain a 4-sized signal_arr 
=> **It is here that something's wrong: Going through `smoothing.convolve_weighted()`, the array shape is messed up** (7 for both signal_arr and weights_arr, when we expect them to be 4-shaped), leading to this "index out of range" error described abovehead
<br>

*Possible fixes:*
My understanding of these "smoothing things" is too limited to understand the precise origin of this bug, but here are some ideas:
- In `smoothing.savgol()`, you have the following condition
https://github.com/etal/cnvkit/blob/b218280edb266788ea8326596d4283183ce425ea/cnvlib/smoothing.py#L174
=> Including case where `len(x)==2` should not have a huge impact on segmentation ?
- I don't know where in the code regions from the ".cnr" with depth at 0 are excluded. But maybe you should insert a cutoff there, to exclude regions having a depth bellow a certain value ?
<br>

Thanks in advance for your help.
Best regards.
Felix.

tskir:
Closes #587. Thank you @tetedange13 for an amazingly detailed investigation which helped a lot!

In Savitzky-Golay smoothing, for certain combinations of parameters, a (wing-padded) signal length can turn out to be shorter than the smoothing window requested, which is mathematically impossible to compute and causes an index mismatch exception downstream.

This change avoids this by checking for this condition and reducing the smoothing window length and, if necessary, also the polynomial order. This ensures (compared to e.g. returning the original signal) that at least some sensible smoothing is still being performed.

### DIFF
--- a/cnvlib/smoothing.py
+++ b/cnvlib/smoothing.py
@@ -58,6 +58,19 @@ def _pad_array(x, wing):
                            x[:-wing-1:-1]))
 
 
+def _check_smoothing_parameters(signal, window_width, order):
+    """Checks and, if necessary, corrects the smoothing parameters. If the signal is shorter than a window width, shrink
+    the window down to the nearest odd number. Additionally, if necessary, shrink the polynomial order down to one-half
+    of the adjusted window width."""
+    if len(signal) < window_width:
+        logging.warning('Convolution window width {} is larger than the padded signal length {}.'.format(
+            window_width, len(signal)))
+        window_width = len(signal) if len(signal) % 2 == 1 else len(signal) - 1
+        order = min(order, window_width // 2)
+        logging.warning('Reduced window to {} and polynomial order to {}'.format(window_width, order))
+    return window_width, order
+
+
 def rolling_median(x, width):
     """Rolling median with mirrored edges."""
     x, wing, signal = check_inputs(x, width)
@@ -185,24 +198,14 @@ def savgol(x, total_width=None, weights=None,
     if weights is None:
         x, total_wing, signal = check_inputs(x, total_width, False)
         y = signal
-        if len(y) < window_width:
-            logging.warning('Convolution window width {} is larger than the padded signal length {}.'.format(
-                window_width, len(y)))
-            window_width = len(y) if len(y) % 2 == 1 else len(y) - 1
-            order = min(order, window_width // 2)
-            logging.warning('Reduced window to {} and polynomial order to {}'.format(window_width, order))
+        window_width, order = _check_smoothing_parameters(y, window_width, order)
         for _i in range(n_iter):
             y = savgol_filter(y, window_width, order, mode='interp')
         # y = convolve_unweighted(window, signal, wing)
     else:
         # TODO fit edges here, too
         x, total_wing, signal, weights = check_inputs(x, total_width, False, weights)
-        if len(signal) < window_width:
-            logging.warning(('Convolution window width {} is larger than the padded signal length {} and will be '
-                             'reduced accordingly.').format(window_width, len(signal)))
-            window_width = len(signal) if len(signal) % 2 == 1 else len(signal) - 1
-            order = min(order, window_width // 2)
-            logging.warning('Reduced window to {} and polynomial order to {}'.format(window_width, order))
+        window_width, order = _check_smoothing_parameters(signal, window_width, order)
         window = savgol_coeffs(window_width, order)
         y, w = convolve_weighted(window, signal, weights, n_iter)
     # Safety
--- a/cnvlib/smoothing.py
+++ b/cnvlib/smoothing.py
@@ -58,19 +58,6 @@ def _pad_array(x, wing):
                            x[:-wing-1:-1]))
 
 
-def check_smoothing_parameters(signal, window_width, order):
-    """Check and, if necessary, correct the smoothing parameters. If the signal is shorter than a window width, shrink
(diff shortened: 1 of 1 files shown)


tskir:
@etal I'm reasonably certain that this should be a sound solution (and I also verified that it definitely fixes #587), but perhaps it would be useful if you could double-check that the math checks out, as it's not quite my area of expertise

etal:
Interesting, I think I meant for `check_inputs` to do this operation internally, but apparently it doesn't.

Would you like to attempt merging `check_inputs` and `check_smoothing_parameters` into a single function? If not, and you're happy with the results of this function, then I'm fine with merging this PR as it is.

tetedange13:
This PR should also fix a tinu bug I got from running `cnvkit.py scatter --by-bin --trend my_patient.cnr`, where chrY crashes due to `--trend` flag

Full traceback here: [traceback.txt](https://github.com/etal/cnvkit/files/6365434/traceback.txt)
=> It is a bit different from #587, IMHO core problem is the same

Here is a minimal".cnr" that reproduce the bug: [sample.cnr.txt](https://github.com/etal/cnvkit/files/6365485/sample.cnr.txt)
<br>

Hope this helps.
Thanks again.
Felix.

tskir:
@etal Thank you for your comments! I've now investigated this further and here's what I found.

You are right: `check_inputs` already checks for this condition [via `_width2wing`](https://github.com/etal/cnvkit/blob/9dd1e7c83705d1e1de6e6e4ab9fdc6973bf4002f/cnvlib/smoothing.py#L41-L44), and for short signals the wing size is correctly reduced when necessary.

The problem is how these results are then used. After calling `check_inputs`, all functions _except_ `savgol()` recalculate the final window width as `2 * total_wing + 1`, which will always work.

Becuase `savgol()` was not taking that into account, this could potentially cause _two_ distinct problems:
* For short signals, the originally requested `total_width` is too big;
* And for _really_ short signals, even the one-iteration window width (7 by default) is too big.

I have now rewritten this PR to properly address both of them. An additional function is no longer required, as well as any modifications to `check_inputs`.

tskir:
@tetedange13 Thank you for the additional test file! I've verified that the latest version of this PR should indeed fix the `scatter` crash.

etal (approved):
This is good. Thanks!

Issue #304, opened 2018-01-22
coverage: process targets and antitargets together


etal:
The `coverage` command currently processes target and antitarget BED files separately (using pysam's bedcov, which does a separate pileup in each BED region).

The mosdepth algorithm is much faster, but works best by scanning a whole chromosome in one shot (I think). Calculating target and antitarget coverages separately would take nearly twice as long if handled separately, right?

Change the `coverage` arguments/usage so that it's possible to feed the command both the targets and antitargets at once. Then it will be straightforward to combine these files and feed them to mosdepth. Also consider that `coverage` will generate as many files, and `cnvlib.do_coverage` will return as many values, as there were BED files given as input.

PR #598, opened 2021-04-24
Fix coverage calculation using counts

### LINKED ISSUE #593: Coverage error
Hi there, I'm executing the cnvkit pipeline by running each command (not batch mode) on[ this dataset](https://console.cloud.google.com/storage/browser/brain-genomics-public/research/sequencing/fastq/novaseq/wes_agilent/75x;tab=objects?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false) (in particular samples HG002, HG003, HG004) from [this paper](https://www.biorxiv.org/content/10.1101/2020.12.11.422022v1.full) with this [target file ](https://kb.10xgenomics.com/hc/en-us/articles/115004150923-Where-can-I-find-the-Agilent-Target-BED-files-) Agilent v7 rgion file.

ref= hg38
blacklist=[hg38-blacklist.v2.bed.gz](https://github.com/Boyle-Lab/Blacklist/tree/master/lists)

Specifically I run
`cnvkit.py target {input.target} --annotate {input.assembly-gtf} -a {params.avg_size} --split -o {output.target} --short-names ` 

`cnvkit.py access {input.ref_fa} -o {output.accessibility} -x {input.blacklist}`

`cnvkit.py antitarget {input.target} -g {input.accessibility} -o {output.antitarget} -a {params.avg_size}`

`cnvkit.py autobin {input.bam_samples} -t {input.target} -g {input.accessibility} -m {params.method}`

`cnvkit.py coverage  {input.bam_sample} {input.target} -p {threads} -o {output.sample_target}  -c -q 10 && \
 cnvkit.py coverage {input.bam_sample} {input.antitarget} -p {threads} -o {output.sample_antitarget} -c -q 10`

And Here I get the error in Coverage step. According to the log in coverage.py file when it tryes to calculate log2 depth it crashes. In the file it's written that log2 depth is `math.log(depth,2) if depth else NULL_LOG2_COVERAGE` so 
when depth is 0 `NULL_LOG2_COVERAGE` is taken
when depth is > 0 depth is taken
but when depth is < 0 an error is thrown `math domain error`

I have thus 2 questions:
1) h**ow can it be possible that a sample bam file has depth < 0 on target file created with also this bam file (aka autobin step)?**
2) **is it possible to include an exception to handle negative depth values?**

Thank you very much

tskir:
Fixes #593. Thank you @joys8998 for a super detailed bug report and @tetedange13 for helping out!

This has to do with the alternative coverage calculation method (`cnvkit.py coverage -c`). For reads with insertions/deletions, calculations using `rlen` and `read.pos` will produce incorrect results (for example, as reported in the linked ticket, a negative total read depth).

Instead, for a general solution, we can use the `.positions` attribute provided by pysam to see where each base maps to the reference genome, and count only the bases which come from within the specified region.

### DIFF
--- a/cnvlib/coverage.py
+++ b/cnvlib/coverage.py
@@ -139,12 +139,7 @@ def region_depth_count(bamfile, chrom, start, end, gene, min_mapq):
         if filter_read(read):
             count += 1
             # Only count the bases aligned to the region
-            rlen = read.query_alignment_length
-            if read.pos < start:
-                rlen -= start - read.pos
-            if read.pos + read.query_alignment_length > end:
-                rlen -= read.pos + read.query_alignment_length - end
-            bases += rlen
+            bases += sum([1 for p in read.positions if start <= p < end])
     depth = bases / (end - start) if end > start else 0
     row = (chrom, start, end, gene,
            math.log(depth, 2) if depth else NULL_LOG2_COVERAGE,


tskir:
I have verified that this indeed resolves the issue. It also runs in reasonable time, only about twice as long as the the standard `cnvkit.py coverage`, and produces reasonable results.

What's interesting, however, is that `coverage -c` doesn't always agree with `coverage`, sometimes quite drastically (excuse the hastily built Google Sheets scatterplot):
image

X axis is the standard method and Y axis is with `-c`. The values are log2.

I suppose it is expected that the two methods sometimes produce different results, but perhaps this is something to be investigated in the future.

tskir:
And here's another quick graph; this is comparing `coverage -c` before (X) and after (Y) the fix:

image

(In "before the fix", the regions with negative absolute width are converted to log2 = –20.)

This is just to verify that the fix didn't drastically break anything. As we can see, most of the regions are right where they were, with minor adjustments in cases where were insertions/deletions affecting the by-count calculation.

etal:
The `-c` flag is there for exploring different coverage calculation strategies, basically. It makes sense that `-c` could under-count regions with gapped or secondary alignments, or other quirks.

Here's the rationale behind the two `coverage` methods:
* Default: Use htslib's pileup engine to calculate per-base coverages exactly the way `samtools depth` would to it. CNVkit does very little of its own math here. It would take a strong argument to justify doing coverage calculations any other way.
* `-c`: Some other copy number tools take the more naive and possibly faster approach of just counting read start positions, or midpoints, within each bin. It's a little tricky to avoid double-counting reads that span bin boundaries. In this method, CNVkit attempts to behave like other tools, with a bit of extra logic to discount the bases that span bin boundaries.

At the time it was released, `-c` was about twice as fast as the default, but less robust and accurate than the pileup engine. Since then, the pileup engine has probably gotten faster, too. So, I would not recommend using `-c` in production, only for algorithm exploration.

Using `read.positions` makes sense, and I'm surprised the result still isn't identical to the default pileup engine's. Maybe secondary reads are being handled differently.

Another method that would be interesting to implement here, and hopefully significantly faster, is mosdepth: #304

**Summary:** This fix looks good to me, and I don't mind the performance hit.

etal (approved):

PR #603, opened 2021-04-26
Update the region regex to allow NCBI-style chromosome names

### LINKED ISSUE #602: unable to 'scatter' plot region of chromosome name with punctuation marks
Hello, 

I am trying to plot a region of a plant chromosome but to no avail. I first plotted successfully the whole chromosome `NC_039902.1` of my sample `1-C7` using the following command:
```
cnvkit.py scatter 1-C7.cnr \
-s 1-C7.cns \
-c NC_039902.1 \
-o 1-C7-NC_039902.1.png
```
When I want to zoom into a specific region of the chromosome, I get the following error:
```
ValueError: Invalid range spec: NC_039902.1:10000000-15000000 (should be like: chr1:2333000-2444000)
```
The command I am using for plotting the region wanted is the following:
```
cnvkit.py scatter 1-C7.cnr \
-s 1-C7.cns \
-c NC_039902.1:10000000-15000000 \
-o 1-C7-NC_039902.1-zoomed.png
```
It seems that when I provide a region in `-c NC_039902.1:10000000-15000000`, the CNVkit raises an error due to the period or point `.` in the chromosome name `NC_039902.1`. The error is not raised when I delete the chromosome name suffix `.1`, but obviously the output figure is empty since the specific chromosome name `NC_039902` does not exist in my bed/fasta/bam files.
 
I am using the original NCBI chromosome names and wish to keep them unchanged. Is there a way around this issue for the cnvkit.py scatter command to tolerate punctuation marks in the chromosome names? Some of the chromosome names I am using are: NC_039898.1 NC_039899.1 NC_039900.1 NC_039901.1 NC_039902.1 etc. And not: chr1 chr2 chr3 etc.

Thanks ahead of time!

-Anibal

tskir:
Closes #602.

The existing regex for the genomic range does not allow NCBI-style chromosome names, such as NC_039898.1.

After this update, it will still require that the chromosome name starts with an alphanumeric character; however, in subsequent characters dots are now allowed.

### DIFF
--- a/skgenome/rangelabel.py
+++ b/skgenome/rangelabel.py
@@ -11,7 +11,7 @@ import re
 Region = collections.namedtuple('Region', 'chromosome start end')
 NamedRegion = collections.namedtuple('NamedRegion', 'chromosome start end gene')
 
-re_label = re.compile(r'(\w+)?:(\d+)?-(\d+)?\s*(\S+)?')
+re_label = re.compile(r'(\w[\w.]*)?:(\d+)?-(\d+)?\s*(\S+)?')
 
 
 def from_label(text, keep_gene=True):


etal (approved):
Subtle. LGTM. Thanks!

PR #608, opened 2021-04-29
Allow specifying output filenames in autobin

### LINKED ISSUE #607: autobin command doesn't include output parameter
Hi,
I just want to add an suggestion. 

I am creating pipeline via snakemake. While I integrating CNVkit into pipeline, autobin step give "output" error because of there is no parmeter as -o/--output

I would be very glad if the output parameter is added to this command. Thank you!

tskir:
Closes #607.

The arguments are optional; if they are not specified, then the behaviour of the command will be exactly the same as before the changes.


etal (approved):
LGTM, thanks!

PR #613, opened 2021-05-09
Fix median-size file selection

### LINKED ISSUE #611: midsize_file Function Does Not Return Median File
From autobin.py:

```
def midsize_file(fnames):
    """Select the median-size file from several given filenames.

    If an even number of files is given, selects the file just below the median.
    """
    return sorted(fnames, key=lambda f: os.stat(f).st_size
                 )[len(fnames) // 2 - 1]
```

- Consider 8 files. The median should be the 4th file in this case (median is between 4/5), or list index 3. This is correctly calculated
```
8 // 2 = 4
4 - 1 = 3
```

- Consider 7 files. The median should be the 4th file, or list index 3. This is incorrectly calculated as index 2, the value just below the median
```
7 // 2 = 3
3 - 1 = 2
```

tskir:
Fixes #611, as described in the issue. Now this will work for both odd and even numbers of filenames provided. Also added an assert for the case if no filenames are provided.

### DIFF
--- a/cnvlib/autobin.py
+++ b/cnvlib/autobin.py
@@ -17,8 +17,8 @@ def midsize_file(fnames):
 
     If an even number of files is given, selects the file just below the median.
     """
-    return sorted(fnames, key=lambda f: os.stat(f).st_size
-                 )[len(fnames) // 2 - 1]
+    assert fnames, 'No files provided to calculate the median size.'
+    return sorted(fnames, key=lambda f: os.stat(f).st_size)[(len(fnames) - 1) // 2]
 
 
 def do_autobin(bam_fname, method, targets=None, access=None,


cnvlib/autobin.py
```diff
@@ -17,8 +17,8 @@ def midsize_file(fnames):
 
     If an even number of files is given, selects the file just below the median.
     """
-    return sorted(fnames, key=lambda f: os.stat(f).st_size
-                 )[len(fnames) // 2 - 1]
+    assert fnames, 'No files provided to calculate the median size.'
```
etal, line 20: I think this should be a ValueError instead of an assert, since the reported problem would be with the function's input, not with the logic or internal consistency of the function itself.

Thanks for fixing the math bug.

etal (changes requested):

etal (approved):

PR #623, opened 2021-06-03
Experimental support for GISTIC export

### LINKED ISSUE #622: How to export gistic markers?
Within `export.py` there's a `export_gistic_markers(cnr_fnames)`.
Also within `export.py`, there's a dictionary...
```
EXPORT_FORMATS = {
    'cdt': fmt_cdt,
    # 'gct': fmt_gct,
    'gistic': export_gistic_markers,
    'jtv': fmt_jtv,
    'nexus-basic': export_nexus_basic,
    'nexus-ogt': export_nexus_ogt,
    'seg': export_seg,
    'theta': export_theta,
    'vcf': export_vcf,
}
```

I've tried exporting a gistic marker file using `cnvkit.py export gistic` but I get an error...
```
# cnvkit.py export gistic
usage: cnvkit.py export [-h] {bed,seg,vcf,theta,nexus-basic,nexus-ogt,cdt,jtv} ...
cnvkit.py export: error: invalid choice: 'gistic' (choose from 'bed', 'seg', 'vcf', 'theta', 'nexus-basic', 'nexus-ogt', 'cdt', 'jtv')
```

**What is the usage to get a gistic marker output file from cnvkit?**

tetedange13:
Should fix issue #622 
=> Simply copy-pasted (+ adapted) code for 'jtv' export (same case with multiple ".cnr" as input) to enable 'gistic' export through command-line
=> This supposes "core" function for "gistic" export (within `cnvlib/export.py`) is correct, which I have not test
=> Maybe further tests are required ?

Closes #622

### DIFF
--- a/cnvlib/commands.py
+++ b/cnvlib/commands.py
@@ -1889,6 +1889,20 @@ P_export_jtv.add_argument('-o', '--output', metavar="FILENAME",
         help="Output file name.")
 P_export_jtv.set_defaults(func=_cmd_export_jtv)
 
+def _cmd_export_gistic(args):
+    formatter = export.EXPORT_FORMATS['gistic']
+    outdf = formatter(args.filenames)
+    write_dataframe(args.output, outdf)
+    
+P_export_gistic = P_export_subparsers.add_parser('gistic',
+                                              help=_cmd_export_gistic.__doc__)
+P_export_gistic.add_argument('filenames', nargs='+',
+        help="""Log2 copy ratio data file(s) (*.cnr), the output of the
+                'fix' sub-command.""")
+P_export_gistic.add_argument('-o', '--output', metavar="FILENAME",
+        help="Output file name.")
+P_export_gistic.set_defaults(func=_cmd_export_gistic)
+
 
 # version ---------------------------------------------------------------------
 
--- a/cnvlib/export.py
+++ b/cnvlib/export.py
@@ -369,7 +369,7 @@ def export_gistic_markers(cnr_fnames):
     #   detect duplicates & exclude them
     # seen_marker_ids = None
     for fname in cnr_fnames:
-        cna = read_cna(fname).autosomes()
+        cna = read_cna(fname)
         marker_ids = cna.labels()
         tbl = pd.concat([
             pd.DataFrame({
--- a/doc/importexport.rst
+++ b/doc/importexport.rst
@@ -146,7 +146,9 @@ cdt, jtv
 ````````
 
 A collection of probe-level copy ratio files (``*.cnr``) can be exported to Java
-TreeView via the standard CDT format or a plain text table::
+TreeView via the standard CDT format or a plain text table:
+
+::
 
     cnvkit.py export jtv *.cnr -o Samples-JTV.txt
     cnvkit.py export cdt *.cnr -o Samples.cdt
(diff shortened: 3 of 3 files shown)


etal:
Looks good, thanks. I think I left this feature as hidden because I hadn't tested its results with Gistic myself. Are you in a position to check whether the output of this command works with Gistic without further modification?

tetedange13:
Hi @etal ,
<br>

Sorry, I have never used GISTIC and I am completely unfamiliar with it
=> However I installed it and as far as I understood it takes 2 input files that can be both produced by `cnvkit.py export` 
- A mandatory "seg" file, whose format matches output from `cnvkit.py export seg <SEVERAL_CNS>` (I mean GISTIC is taking it without error)
- An optionnal "markers" file which is produced by `cnvkit.py export gisitic <SEVERAL_CNR>`

Feeding GISTIC **only** with mandatory "seg" file works fine, but I got an **issue feeding it with both "seg" and "markers" files**:
```bash
GISTIC 2.0 input error detected:
74 segment start or end positions in 'cns.gistic.seg' do not match any markers in 'cnr.gistic.markers'.
First bad position is X:150501 at line 114.
```
=> I guess it is because `cnvkit.py export gistic` [works only on autosomes](https://github.com/etal/cnvkit/blob/25d91cdb962639c98d041c3b05c7de659c41dc70/cnvlib/export.py#L372) (contrary to `cnvkit.py export seg`)
=> Fortunately, filtering out sexual chromosomes from "seg" file produced by CNVkit is enough for GISTIC to finish without error !

To conclude: this seems OK to me, but as I said I do not know enough about GISTIC to assert results generated from "GISTIC + CNVkit's exports" are accurate
=> Maybe you should rather wait for `BioComSoftware` feedback ?
<br>

 Hope this helps.
Have a nice day.
Felix !

tskir:
@tetedange13 Thank you for the PR Felix! I agree we should ideally wait for @BioComSoftware's feedback prior to merging this.

I also think there are two additional things to be done here:
* Adding some basic documentation about this format to the export section;
* Since GISTIC presumably can only work with autosomes, the SEG export command should have a flag to only export autosomes, to avoid the mismatch which you encountered in your tests.

tetedange13:
Hi @tskir,
<br>

> Since GISTIC presumably can only work with autosomes

Just to be clear, GISTIC itself can deal with sex-chrom (even if they are removed by default according to [doc](http://portals.broadinstitute.org/cgi-bin/cancer/publications/pub_paper.cgi?mode=view&paper_id=216&p=t) and that I could not manage to make it use keep them with `-rx=0` param...)
=> The issue I was getting comes from inconsistency between `cnvkit.py export seg` that includes sex-chrom and current `cnvkit.py export gistic` that removes them internally (based on [here](https://github.com/etal/cnvkit/blob/25d91cdb962639c98d041c3b05c7de659c41dc70/cnvlib/export.py#L372))
=> I do not know if there was a specific reason for that, at the time `export_gistic_markers()` was written?

To sum up:
=> If you want I can commit a sex-chrom switch for `cnvkit.py export seg` 
=> But would not it be better to simply remove the `.autosome()` filter from `export_gistic_markers()` ? (I tested, GISTIC is taking both files just fine)

<br>

Kind regards.
Felix.

tskir:
Based on the feedback we received, I think this would be the most reasonable set of changes:
* Remove the `.autosome()` filter
* Add minimal description of gistic export to the [documentation](https://github.com/etal/cnvkit/blob/master/doc/importexport.rst) and mark it as experimental

@tetedange13 Do you think you could add those changes to this PR?

tetedange13:
Hi @tskir ,

Done ! To document GISTIC export I did my best based on code docstrings + GISTIC documentation

Have a nice day.
Felix.

tskir (approved):
@tetedange13 Thank you Felix, you're a star!

etal (approved):
LGTM, thanks! The `.autosomes()` filter was there because I was under the mistaken impression that GISTIC only accepts autosomes as input.

PR #625, opened 2021-06-04
Support for vertical heatmap and various plotting refactor

### LINKED ISSUE #35: heatmap: Option to transpose the plot
The current "horizontal" layout shows samples in rows, chromosomes in columns. A transposed or "vertical" layout would show samples in columns, chromosomes in rows. (Sometimes this makes a better figure.)

Consider the flags "-t/--transpose" or "-v/--vertical".

### LINKED ISSUE #610: Heatmap borders per sample
Hello, 

I run CNVkit on targeted sequencing data and generated a CN heatmap. I plotted approximately 90 samples in the heatmap which would be very useful to visualize having lines as borders separating the data per sample across all chromosomes. It's not easy to see where the borders of each sample are when plotting many samples on the same heatmap. Is there currently an option not obvious to me doing what I am describing? Please advise.

Many thanks, 

Katerina

tetedange13:
Joint work by @tetedange13 and @tskir:
* Support for vertical heatmap (`--vertical`)
* Support for delimiting samples (`--delimit-samples`)
* Refactor the use of the colourmap
* Various additional refactor

Closes #35. Closes #610.

### DIFF
--- a/cnvlib/commands.py
+++ b/cnvlib/commands.py
@@ -1045,7 +1045,8 @@ def _cmd_heatmap(args):
                                                  args.male_reference)
             cnarr = cnarr.shift_xx(args.male_reference, is_sample_female)
         cnarrs.append(cnarr)
-    heatmap.do_heatmap(cnarrs, args.chromosome, args.desaturate, args.by_bin, args.vertical)
+    heatmap.do_heatmap(cnarrs, args.chromosome, args.desaturate, args.by_bin, 
+                       args.delim_sampl, args.vertical)
     if args.output:
         oformat = os.path.splitext(args.output)[-1].replace(".", "")
         pyplot.savefig(args.output, format=oformat, bbox_inches="tight")
@@ -1073,6 +1074,9 @@ P_heatmap.add_argument('-d', '--desaturate', action='store_true',
         help="Tweak color saturation to focus on significant changes.")
 P_heatmap.add_argument('-v', '--vertical', action='store_true',
         help="Plot heatmap with samples as X-axis (instead of Y-axis).")
+P_heatmap.add_argument('--delimit-samples',
+        action='store_true', dest='delim_sampl',
+        help="Add an horizontal delimitation line between each sample.")
 P_heatmap.add_argument('-y', '--male-reference', '--haploid-x-reference',
         action='store_true',
         help="""Assume inputs were normalized to a male reference
--- a/cnvlib/commands.py
+++ b/cnvlib/commands.py
@@ -1045,7 +1045,7 @@ def _cmd_heatmap(args):
                                                  args.male_reference)
             cnarr = cnarr.shift_xx(args.male_reference, is_sample_female)
         cnarrs.append(cnarr)
-    heatmap.do_heatmap(cnarrs, args.chromosome, args.desaturate, args.by_bin)
+    heatmap.do_heatmap(cnarrs, args.chromosome, args.desaturate, args.by_bin, args.vertical)
     if args.output:
         oformat = os.path.splitext(args.output)[-1].replace(".", "")
         pyplot.savefig(args.output, format=oformat, bbox_inches="tight")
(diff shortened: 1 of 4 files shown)


tskir:
Thank you Felix for implementing this long wanted feature! I especially like the idea of using a colormap, as opposed to coloring each rectangle separately.

The only concern I have is dropping the desaturate option, which can be quite useful at times. However, I found a way to revive it. In fact, your work has inspired me to do some additional refactor of this module. I'll be submitting a companion PR shortly

tetedange13:
Last commit, made from @tskir modifcations should close #610.
=> If prefered, I can also make a dedicated PR as soon as changes from #632 are merged

Kind regards.
Felix.

tetedange13:
Hi @tskir ,

I saw you renamed this PR for "no review required"
=> I do not know if you noticed, but I added a commit on top of your modifications, to implement "sample delimitation"

What do you think, is it better if I close this PR then open another as soon as yours are merged, to add the commits implementing "sample delimitation" ?

Thanks for your answer.
Have a nice day.
Felix.

tskir:
@tetedange13 Thank you for pointing this out! I indeed didn't notice that last commit. As all changes from #632 are already included here, I've closed that one. So now this is the only PR left to be reviewed & merged

tskir (approved):

etal (approved):
Great new features, thanks! This was a gnarly part of the code base.

Are there any performance implications? Is it noticeably faster (or slower) with these different matplotlib functions?

tetedange13:
Hi @etal,
<br>

I have not done extensive performance tests and I do not have very huge data (small hybrid-capture panel)
=> What I can tell you for sure about my part is that, when I first tried to implement "vertical heatmap" keeping up with `BrokenBarHCollection()`, verticalized plot was quite heavy to plot for Matplotlib (loading all the time, zooming locked...)

Then @tskir's smart implementation of `do_desaturate` seems more algorithmically efficient (transforming "linspaced values" only, instead of whole dataframe), so I guess it will bring a bit of speed ?
https://github.com/etal/cnvkit/blob/cb3e6fd82ff24091c141f1e8ca3647949f12953f/cnvlib/heatmap.py#L162
<br>

My pleasure to help !
Have a nice day.
Felix.

tetedange13:
I noticed `cna2df()` small function still has a `do_desature` param which is synced with `args.do_desaturate`:
https://github.com/etal/cnvkit/blob/cb3e6fd82ff24091c141f1e8ca3647949f12953f/cnvlib/heatmap.py#L77-L84
=> I guess it was not in @tskir intentions, as **"desature" process is [kind of done twice](https://github.com/etal/cnvkit/blob/cb3e6fd82ff24091c141f1e8ca3647949f12953f/cnvlib/heatmap.py#L162) now)** ?
=> Plus this function can probably be factorized (or deleted), because now it only gets 3 columns from `cnarr.data`
(something like: `cna2df = lambda cna: cna.data.loc[:, ['start', 'end', 'log2']]`)

Do you want me to submit a new separate PR ?
=> I also wanted to add an optional `heatmap --title` parameter, so I could do both
<br>

Hope this helps.
Felix.

etal:
I see the second do_desaturate, but isn't that just for the legend on the side?

A `--title` parameter would be nice, to match the `scatter` command. That could be a new PR if you'd like.

tetedange13:
Hum I do not think so, with `pcolor()` legend colorbar is derived from Matplotlib object returned:
https://github.com/etal/cnvkit/blob/a402d901856fa2b72489c42aa5e1b0a975a730f0/cnvlib/heatmap.py#L163-L164

Actually `do_desaturate` arg is applied in `cna2df` only to create a new 'color' column
https://github.com/etal/cnvkit/blob/a402d901856fa2b72489c42aa5e1b0a975a730f0/cnvlib/heatmap.py#L17
=> **But as this 'color' column is not used anymore in the code, this transformation step has no impact on final output (but probably a bit on performances)**
=> Maybe we can wait for @tskir's confirmation, but IMO these are useless lines of code

I also tested with above-mentionned `cna2df = lambda` instead and it works like a charm 
<br>

OK I will submit a new PR !
Have a nice day.
Felix.

PR #638, opened 2021-07-02
Improve error message when coordinate set contains duplicate values

### LINKED ISSUE #637: cnvkit fix doesn't produce cnr file
Hi,

So I originally ran the batch command on my samples and everything was processed without any error messages. It seemed to have stopped working correctly at the cnvkit fix stage as the target and antitarget files ares produced for each bam file.

So I decided to run the fix command on a set of antitarget and target files. Again there was no output .cnr file produced, and there was no error message. I used the following cnvkit command, and also included some of the what was printed to the terminal while it ran (positions from the ref file). Any idea why no cnr file is produced?

`cnvkit.py fix results/SRR5206578_filt2.targetcoverage.cnn results/SRR5206578_filt2.antitargetcoverage.cnn /mnt/DATA5/tmp/kev/tmpDbs/SRA/iKapMice2/FlatReference.cnn -o results/SRR5206578.cnr`

`('chrX', 134336084, 134336174)
('chrX', 134336084, 134336174)
('chrX', 134336510, 134336639)
('chrX', 134340347, 134340412)
('chrX', 134340347, 134340412)
('chrX', 134340347, 134340412)`

tskir:
Closes #637.

Previously, when the coordinate set contains lots of duplicate values
(on the order of thousands and more), the exception would format and
dump all of them to stderr, obscuring the initial error message.

Now, only the number of duplicated regions is output, with up to ten
first regions, making the error more readable in all cases.

### DIFF
--- a/cnvlib/fix.py
+++ b/cnvlib/fix.py
@@ -135,8 +135,9 @@ def match_ref_to_sample(ref_cnarr, samp_cnarr):
                        (ref_labeled, "reference")):
         dupes = dset.index.duplicated()
         if dupes.any():
-            raise ValueError("Duplicated genomic coordinates in " + name +
-                             " set:\n" + "\n".join(map(str, dset.index[dupes])))
+            raise ValueError(
+                "Duplicated genomic coordinates in {} set. Total duplicated regions: {}. First ten regions: {}.".format(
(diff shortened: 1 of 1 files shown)


tskir:
The updated error message looks like this:

image

etal (approved):
Much better, thanks.

PR #641, opened 2021-07-06
Parameter to set 'scatter' plot size

### LINKED ISSUE #600: Modify width/height for output figure generated from scatter command
Hi,

As cnvkit provide a great convenience to analyze the CNV status of the sample.
One can generate a CNV call plot from the cns and cnr data.

I am wondering if I can give a option to set the width/height/resolution of the output png figure, for either whole genome or selected region/chr?

Thanks,
Junfeng

tetedange13:
Should close #600.

Hope this helps.
Kind regards.
Felix.

### DIFF
--- a/cnvlib/commands.py
+++ b/cnvlib/commands.py
@@ -928,7 +928,7 @@ def _cmd_scatter(args):
         ("window_width", args.width),
         ("y_min", args.y_min),
         ("y_max", args.y_max),
-        ("plot_dim", args.plot_dim),
+        ("fig_size", args.fig_size),
         ("antitarget_marker", args.antitarget_marker),
         ("segment_color", args.segment_color),
     ) if v is not None}
@@ -1007,7 +1007,8 @@ P_scatter_aes.add_argument('-t', '--trend', action='store_true',
         help="Draw a smoothed local trendline on the scatter plot.")
 P_scatter_aes.add_argument('--y-max', type=float, help="y-axis upper limit.")
 P_scatter_aes.add_argument('--y-min', type=float, help="y-axis lower limit.")
-P_scatter_aes.add_argument('-p', '--plot-dim', nargs=2, type=float,
+P_scatter_aes.add_argument('--fig-size', nargs=2, 
+        metavar=('WIDTH','HEIGHT'), type=float,
         help="""Width and height of the plot in inches.
                 [Default: Pre-defined in Matplotlib 'rcParams' variable
                 (most of the time: '6.4 4.8')]""")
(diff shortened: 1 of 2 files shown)


cnvlib/commands.py
```diff
@@ -1006,6 +1007,10 @@ def _cmd_scatter(args):
         help="Draw a smoothed local trendline on the scatter plot.")
 P_scatter_aes.add_argument('--y-max', type=float, help="y-axis upper limit.")
 P_scatter_aes.add_argument('--y-min', type=float, help="y-axis lower limit.")
+P_scatter_aes.add_argument('-p', '--plot-dim', nargs=2, type=int,
```
tskir, line 1010:
```suggestion
P_scatter_aes.add_argument('-p', '--plot-dim', nargs=2, type=float,
```

Since dimentions can be float in MPL

cnvlib/commands.py
```diff
@@ -1006,6 +1007,10 @@ def _cmd_scatter(args):
...
+        help="""Dimensions of plot, such as a list of 'width height' in inches.
+                [Default: Pre-defined in Matplotlib 'rcParams' variable
+                (most of the time [6.4,4.8]]""")
```
tskir, line 1013:
```suggestion
        help="""Width and height of the plot in inches.
                [Default: Pre-defined in Matplotlib 'rcParams' variable
                (most of the time: '6.4 4.8')]""")
```
Adjusted the wording slightly

tskir (approved):
Thank you Felix! This can be very useful

cnvlib/commands.py
```diff
@@ -1006,6 +1007,10 @@ def _cmd_scatter(args):
         help="Draw a smoothed local trendline on the scatter plot.")
 P_scatter_aes.add_argument('--y-max', type=float, help="y-axis upper limit.")
 P_scatter_aes.add_argument('--y-min', type=float, help="y-axis lower limit.")
+P_scatter_aes.add_argument('-p', '--plot-dim', nargs=2, type=float,
```
etal, line 1010: How about calling it "figure size" instead of "dim", skipping the short form of the argument (it's not a recurring option in other commands), and setting the metavar for clarity:

    P_scatter_aes.add_argument('--figure-size', nargs=2, metavar=('WIDTH', 'HEIGHT')), type=float,

tetedange13: Hi @etal,

Thanks for your code review ! You are right, here is result "help" of commit 85374df3bbecc8aa9805a1ebe3076c461b0b9441 (+ I changed variable name everywhere else):
```python
  --fig-size WIDTH HEIGHT
                        Width and height of the plot in inches. [Default: Pre-defined in Matplotlib 'rcParams'
                        variable (most of the time: '6.4 4.8')]
```

Have a nice day.
Felix.

etal (approved):
Great, thanks! LGTM now.

asaki1986:
thanks, plotsize indeed help a lot.

However, the scatter point size turned to be relatively small when figure width was set larger than 10.

Any method to increase the dot size a bit?

PR #645, opened 2021-07-12
Triangles for seg masked by pruning at 'ymin=-5'

### LINKED ISSUE #643: Default y-min in chromsome scatter plots
I wonder if this should be `min` instead of `max`?

https://github.com/etal/cnvkit/blob/d4145c5e9b97817c2d625f07b44212c4176b5a1f/cnvlib/scatter.py#L370

With the current default settings, some biallelic deletion are not shown. Here is an example (CNVkit v0.9.9) without `--y-min` specified:

Screen Shot 2021-07-08 at 16 39 44

and with `--y-min=-10`:

Screen Shot 2021-07-08 at 16 39 50

tetedange13:
Should close #643, if this is well an unexpected behaviour (which I guess, but not 100% sure)

Have a nice day.
Felix.

### DIFF
--- a/cnvlib/scatter.py
+++ b/cnvlib/scatter.py
@@ -367,7 +367,7 @@ def cnv_on_chromosome(axis, probes, segments, genes, antitarget_marker=None,
 
     # Configure axes
     if not y_min:
-        y_min = min(-5.0, min(y.min() - .1, -.3)) if len(y) else -1.1
+        y_min = max(-5.0, min(y.min() - .1, -.3)) if len(y) else -1.1
     if not y_max:
         y_max = max(.3, y.max() + (.25 if genes else .1)) if len(y) else 1.1
     if x_limits:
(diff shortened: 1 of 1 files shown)


tskir (changes requested):
No specific changes are requested for now, but this PR is on hold until discussions in #643 finish

tetedange13:
Should close #643.

Default behaviour when using `scatter --chromosome` is to "prune" at 'ymin=-5' (see https://github.com/etal/cnvkit/issues/643#issuecomment-883936000)
=> This behaviour can cause probes to be masked (no concerns for probes) but for segments it can be problematic

With this PR, having masked seg raises a warning + they are signaled as triangles crossing Y-axis
=> X-pos of a triangle is determined by `seg.start`, so that number of seen triangles does not always match number of masked seg, due to 2 masked seg having same `seg.start` (which is the case for "amplicon.{cnr,cns}" - chr9)

See https://github.com/etal/cnvkit/issues/643#issuecomment-890461910 for an example obtained with "amplicon.{cnr,cns}" test files
<br>

Have a nice day.
Felix.

etal (approved):
LGTM, thanks!

PR #646, opened 2021-07-12
Improve bintest doc

### LINKED ISSUE #621: Difference of log2 values between .cnr file and .bintest
Hi,

Why log2 values are different between cnr file and bintest.cns file?

Here is my example:

**- In .cnr file, all bins on BRCA2  have a log2 between 1 and 2.5 suggesting and amplification of the entire gene:**

chromosome|start|end|gene|log2|depth|weight
-- | -- | -- | -- | -- | -- | --
chr13 | 32890559 | 32890670 | BRCA2_ex02 | 1.92172 | 11112.5 | 0.989018
chr13 | 32893193 | 32893468 | BRCA2_ex03 | 2.06781 | 9570 | 0.989823
chr13 | 32899192 | 32899327 | BRCA2_ex04 | 1.29903 | 4212.33 | 0.989634
chr13 | 32900217 | 32900293 | BRCA2_ex05 | 1.04573 | 1308.66 | 0.98301
… | &nbsp; | &nbsp; | &nbsp; | &nbsp; | &nbsp; | &nbsp;
chr13 | 32968805 | 32969076 | BRCA2_ex25 | 1.86698 | 12887.1 | 0.993064
chr13 | 32971014 | 32971187 | BRCA2_ex26 | 1.72835 | 12128.8 | 0.991547
chr13 | 32972278 | 32972595 | BRCA2_ex27 | 1.55005 | 14927.4 | 0.993474
chr13 | 32972595 | 32972913 | BRCA2_ex27 | 1.88532 | 14994.7 | 0.993332

**- .cns file agrees with .cnr file showing a log2 for the whole gene of 1.64:**

chromosome | start | end | gene | log2 | depth | probes | weight | ci_lo | ci_hi
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
chr13 | 30937699 | 32972913 | BRCA2_ex02,BRCA2_ex03,BRCA2_ex04,BRCA2_ex05,BRCA2_ex06,BRCA2_ex07,BRCA2_ex08,BRCA2_ex09,BRCA2_ex10,BRCA2_ex11,BRCA2_ex13,BRCA2_ex14,BRCA2_ex15,BRCA2_ex16,BRCA2_ex17,BRCA2_ex18,BRCA2_ex19,BRCA2_ex20,BRCA2_ex21,BRCA2_ex22,BRCA2_ex23,BRCA2_ex24,BRCA2_ex25,BRCA2_ex26,BRCA2_ex27 | 1.63355 | 7483.63 | 60 | 58.3449 | 1.5957 | 1.66772

**- call.cns  reports a CNV on the BRCA2 gene by counting 7 copies with a mean log2 of 1.6:**

chromosome | start | end | gene | log2 | cn | depth | p_ttest | probes | weight
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
chr13 | 30937699 | 32972913 | BRCA2_ex02_UTR,BRCA2_ex02,BRCA2_ex03,BRCA2_ex04,BRCA2_ex05,BRCA2_ex06,BRCA2_ex07,BRCA2_ex08,BRCA2_ex09,BRCA2_ex10,BRCA2_ex11,BRCA2_ex13,BRCA2_ex14,BRCA2_ex15,BRCA2_ex16,BRCA2_ex17,BRCA2_ex18,BRCA2_ex19,BRCA2_ex20,BRCA2_ex21,BRCA2_ex22,BRCA2_ex23,BRCA2_ex24,BRCA2_ex25,BRCA2_ex26,BRCA2_ex27 | 1.60607 | 7 | 7483.63 | 1.29963e-39 | 60 | 58.3449

**- But in bintest.cns file, for the same bins, log2 are really different, some of them are negative:**
(Initialy those bins  were not reported in bintest file, I modified alpha value to 1 in bintest.py to understand why my bins were not reported as significative.)
 
chr13 | 32890559 | 32890670 | BRCA2_ex02 | 0.288171 | 11112.5 | 0.989018 | 0.0161461
-- | -- | -- | -- | -- | -- | -- | --
chr13 | 32893193 | 32893468 | BRCA2_ex03 | 0.434261 | 9570 | 0.989823 | 8.8159e-05
chr13 | 32899192 | 32899327 | BRCA2_ex04 | -0.334525 | 4212.33 | 0.989634 | 0.00354805
chr13 | 32900217 | 32900293 | BRCA2_ex05 | -0.587826 | 1308.66 | 0.98301 | 3.62753e-05
… | &nbsp; | &nbsp; | &nbsp; | &nbsp; | &nbsp; | &nbsp; | &nbsp;
chr13 | 32968805 | 32969076 | BRCA2_ex25 | 0.233426 | 12887.1 | 0.993064 | 0.014072
chr13 | 32971014 | 32971187 | BRCA2_ex26 | 0.0948008 | 12128.8 | 0.991547 | 0.411222
chr13 | 32972278 | 32972595 | BRCA2_ex27 | -0.0835012 | 14927.4 | 0.993474 | 0.411222
chr13 | 32972595 | 32972913 | BRCA2_ex27 | 0.251767 | 14994.7 | 0.993332 | 0.00656489

Can you help me understand why there is such a difference?

Thanks

Flora

tetedange13:
Hi all,
<br>

Regarding #621 and the need to properly document `bintest`

=> I did my best taking info from docstrings and previous `cnv_ztest.py` paragraph
=> Moved it to a dedicated new section inside `pipeline.rst`, condidering it is a separate step of CNVkit's pipeline (one call CNV with `batch` and if some CNV are missing or higher resolution is needed, one can use `bintest`) 
=> But `pipeline.rst` is already really dense, so maybe a better emplacement can be found ?

***Important:***
I think there is a confusion between:
* Extension of the file produced through `batch` ".bintest.cns"
* Output format produced by `bintest`, which is actually closer to a ".cnr" (regarding coordinates mostly)

=> **Maybe something needs a fix here ?** Anyway documentation I wrote is correct but a bit contradictory if you pay attention

**Of course, feel free to tell me if this correct enough to you, or if I need to refactor / change some parts !**
<br>

Hope this helps.
Have a nice day.
Felix.

---

Closes #621

### DIFF
--- a/doc/pipeline.rst
+++ b/doc/pipeline.rst
@@ -828,3 +828,15 @@ criteria are merged together and the column values are recalculated
 appropriately. Segments on different chromosomes or with different
 allele-specific copy number values will not be merged, even if the total copy
 number is the same.
+
+Extracting focal CNV
+````````````````````
+
+Test each bin in a .cnr file individually for non-neutral copy number.
+Specifically, calculate the probability of a bin's log2 value versus a
+normal distribution with a mean of of 0 and standard deviation
+back-calculated from bin weight. Output another .cnr file with z-test
+probabilities in the additional column "ztest"; drop rows where the
+probability is above the threshold (``--alpha``/``-a``).
+Since CNVkit >= v0.9.7, this post-processing step is included into 
+:ref:`batch` command, producing a 3rd segment file: ".bintest.cns"
--- a/doc/pipeline.rst
+++ b/doc/pipeline.rst
@@ -837,7 +837,7 @@ number is the same.
 
 *New in version 0.9.7*
 
-``bintest`` subcommand replaced additional script ``cnv_ztest.py``, aiming to **call focal bin-level CNV**.
+``bintest`` subcommand replaced additional script ``cnv_ztest.py``, aiming to call focal bin-level CNVs.
 
 Each bin in a .cnr file is tested individually for non-neutral copy number. Specifically, we calculate the probability of a bin's log2 value *versus* a normal distribution with a mean of 0 and standard deviation back-calculated from bin weight. Bin p-values are eventually corrected for multiple hypothesis testing by the Benjamini-Hochberg method.
 
--- a/doc/pipeline.rst
+++ b/doc/pipeline.rst
@@ -832,7 +832,10 @@ number is the same.
 Extracting focal CNV
 ````````````````````
 
-Test each bin in a .cnr file individually for non-neutral copy number.
+*New in version 0.9.7*
+
+The additional script ``cnv_ztest.py`` were replaced by a dedicated ``bintest`` 
+command replace each bin in a .cnr file individually for non-neutral copy number.
 Specifically, calculate the probability of a bin's log2 value versus a
 normal distribution with a mean of of 0 and standard deviation
 back-calculated from bin weight. Output another .cnr file with z-test
@@ -840,3 +843,10 @@ probabilities in the additional column "ztest"; drop rows where the
 probability is above the threshold (``--alpha``/``-a``).
 Since CNVkit >= v0.9.7, this post-processing step is included into 
 :ref:`batch` command, producing a 3rd segment file: ".bintest.cns"
+
+.. note::
+    ``bintest`` can be ran with segments specified (through ``-s file.cns``). 
+    This way, calculations are made comparing to the segment to which the bin belongs, 
+    and otherwise to the whole chromosome
+
+    This can lead to , especially on segments undergoing a CNV
--- a/doc/pipeline.rst
+++ b/doc/pipeline.rst
@@ -844,13 +844,28 @@ number is the same.
 
 *New in version 0.9.7*
 
-``bintest`` subcommand replaced additional script ``cnv_ztest.py``, aiming to call focal bin-level CNVs.
+``bintest`` subcommand replaced additional script ``cnv_ztest.py``, aiming to 
+call focal bin-level CNVs.
 
-Each bin in a .cnr file is tested individually for non-neutral copy number. Specifically, we calculate the probability of a bin's log2 value versus a normal distribution with a mean of 0 and standard deviation back-calculated from bin weight. Bin p-values are eventually corrected for multiple hypothesis testing by the Benjamini-Hochberg method.
+Each bin in a .cnr file is tested individually for non-neutral copy number. 
+Specifically, we calculate the probability of a bin's log2 value versus a normal 
+distribution with a mean of 0 and standard deviation back-calculated from bin 
+weight. Bin p-values are eventually corrected for multiple hypothesis testing by 
+the Benjamini-Hochberg method.
 
-Output is another .cnr with aditional column "p_bintest" corresponding to p-value of test probabilities. Rows considered as not significant, i.e. having probability above the threshold (controlled by ``--alpha``/``-a`` parameter), are dropped.
+Output is another .cnr with aditional column "p_bintest" corresponding to p-value 
+of test probabilities. Rows considered as not significant, i.e. having probability 
+above the threshold (controlled by ``--alpha``/``-a`` parameter), are dropped.
 
-This post-processing step were also included into :ref:`batch` subcommand, where ``bintest`` is run with segment file and on target bins only (equivalent to ``-t, --target`` parameter of ``bintest`` subcommand), producing a third ".cns" file with the suffix ".bintest.cns".
+This post-processing step were also included into :ref:`batch` subcommand, where 
+``bintest`` is run with segment file and on target bins only (equivalent to 
+``-t, --target`` parameter of ``bintest`` subcommand), producing a third ".cns" 
+file with the suffix ".bintest.cns".
 
 .. note::
-    If ``bintest`` is run with ``-s file.cns``, it will try to find additional bin-level alterations, considering alterations present in "file.cns" to be the baseline. In other words, when the .cns file is given, comparisons are made in relation to the segment to which the bin belongs, and otherwise to the whole chromosome. This can lead to (apparently) unexpected log2 values, especially on regions undergoing a CNV.
+    If ``bintest`` is run with ``-s file.cns``, it will try to find additional 
+    bin-level alterations, considering alterations present in "file.cns" to be 
+    the baseline. In other words, when the .cns file is given, comparisons are 
+    made in relation to the segment to which the bin belongs, and otherwise to 
+    the whole chromosome. This can lead to (apparently) unexpected log2 values, 
+    especially on regions undergoing a CNV.
--- a/doc/pipeline.rst
+++ b/doc/pipeline.rst
@@ -64,6 +64,13 @@ The pipeline executed by the ``batch`` command is equivalent to::
     # For each tumor sample...
     cnvkit.py fix Sample.targetcoverage.cnn Sample.antitargetcoverage.cnn my_reference.cnn -o Sample.cnr
     cnvkit.py segment Sample.cnr -o Sample.cns
+    
+    # Post-processing for each tumor sample...
+    cnvkit.py segmetrics Sample.cnr -s Sample.cns --ci --alpha 0.5 --smooth-bootstrap -o Sample.segmetrics.cns.tmp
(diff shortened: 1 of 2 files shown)


doc/pipeline.rst
```diff
@@ -821,10 +821,32 @@ combined:
...
 In each case, adjacent segments with the same value according to the given
 criteria are merged together and the column values are recalculated
 appropriately. Segments on different chromosomes or with different
 allele-specific copy number values will not be merged, even if the total copy
 number is the same.
+
+
+.. _bintest:
+
+``bintest``
```
tskir, line 835: First of all, regarding where to put this: since `bintest` is being called as part of the batch command, I think the best place is here, in `pipeline.rst`, even if it's quite extensive already.

By the way, speaking of the batch, could you also please add bintest to the list of commands above, in the “The pipeline executed by the batch command is equivalent to...” section?

doc/pipeline.rst
```diff
@@ -821,10 +821,32 @@ combined:
...
+-----------
```
tskir, line 836: Second, regarding the question you raised about the formats

> I think there is a confusion between:
>
> * Extension of the file produced through batch ".bintest.cns"
> * Output format produced by bintest, which is actually closer to a ".cnr" (regarding coordinates mostly)

I think this is a really good point, and the fact that the file is currently called `.bintest.cns` is probably a typo. Let's rename it to `.bintest.cnr` as part of this PR as well (as well as any occurrences of `.bintest.cns` in the code which you might find)

doc/pipeline.rst
```diff
@@ -821,10 +821,32 @@ combined:
...
+
+*New in version 0.9.7*
+
+``bintest`` subcommand replaced additional script ``cnv_ztest.py``, aiming to **call focal bin-level CNV**.
```
tskir, line 840:
```suggestion
``bintest`` subcommand replaced additional script ``cnv_ztest.py``, aiming to call focal bin-level CNVs.
```

doc/pipeline.rst
```diff
@@ -821,10 +821,32 @@ combined:
...
+
+Each bin in a .cnr file is tested individually for non-neutral copy number. Specifically, we calculate the probability of a bin's log2 value *versus* a normal distribution with a mean of 0 and standard deviation back-calculated from bin weight. Bin p-values are eventually corrected for multiple hypothesis testing by the Benjamini-Hochberg method.
```
tskir, line 842:
```suggestion
Each bin in a .cnr file is tested individually for non-neutral copy number. Specifically, we calculate the probability of a bin's log2 value versus a normal distribution with a mean of 0 and standard deviation back-calculated from bin weight. Bin p-values are eventually corrected for multiple hypothesis testing by the Benjamini-Hochberg method.
```

doc/pipeline.rst
```diff
@@ -821,10 +821,32 @@ combined:
...
+
+Output is another .cnr with aditional column "p_bintest" corresponding to p-value of test probabilities. Rows considered as **not significant**, *i.e.* having probability above the threshold (controlled by ``--alpha``/``-a`` parameter), are **dropped**.
```
tskir, line 844:
```suggestion
Output is another .cnr with aditional column "p_bintest" corresponding to p-value of test probabilities. Rows considered as not significant, i.e. having probability above the threshold (controlled by ``--alpha``/``-a`` parameter), are dropped.
```

doc/pipeline.rst
```diff
@@ -821,10 +821,32 @@ combined:
...
+
+This post-processing step were also included into 
+:ref:`batch` subcommand, where ``bintest`` is run with segment file and on target bins only (equivalent to ``-t, --target`` parameter of ``bintest`` subcommand), producing a 3rd ".cns" file with ".bintest.cns" suffix.
```
tskir, line 847:
```suggestion
This post-processing step were also included into :ref:`batch` subcommand, where ``bintest`` is run with segment file and on target bins only (equivalent to ``-t, --target`` parameter of ``bintest`` subcommand), producing a third ".cns" file with the suffix ".bintest.cns".
```

doc/pipeline.rst
```diff
@@ -821,10 +821,32 @@ combined:
...
+
+.. note::
+    If ``bintest`` is run with ``-s file.cns``, it will try to find additional bin-level alterations, condidering alterations present in "file.cns" as already known ones. Practically with ".cns" given, calculations are made comparing to the segment to which the bin belongs, and otherwise to the whole chromosome. 
+
+    This can lead to (apparently) unexpected log2 values, especially on regions undergoing a CNV.
```
tskir, line 852:
```suggestion
    If ``bintest`` is run with ``-s file.cns``, it will try to find additional bin-level alterations, considering alterations present in "file.cns" to be the baseline. In other words, when the .cns file is given, comparisons are made in relation to the segment to which the bin belongs, and otherwise to the whole chromosome. This can lead to (apparently) unexpected log2 values, especially on regions undergoing a CNV.
```

tskir (changes requested):
Thank you once again Felix for spearheading this! This would have been another issue probably not resolved for months/years without you. I left a few suggesions, after addressing them we can merge

doc/pipeline.rst
tetedange13: Adding `bintest` to this section, I also added missing steps for ".call.cns" generation and several points regarding [tricky post-processing part](https://github.com/etal/cnvkit/blob/25d91cdb962639c98d041c3b05c7de659c41dc70/cnvlib/batch.py#L195-L211):
- `segmentrics --ci` is run with alpha=0.5 (instead of default=0.05=5% as usually done in statistics). Do you have an idea why ?
- `segmetrics.do_segmetrics()` has no `skip_low=False` flag as argument
=> Possible `--drop-low-coverage` is only applied when running `cnvkit.py segmetrics` alone, [within `commands.py`](https://github.com/etal/cnvkit/blob/25d91cdb962639c98d041c3b05c7de659c41dc70/cnvlib/commands.py#L1324-L1325)
=> Maybe we should add this arg to `segmetrics.do_segmetrics()` and sync it with `batch --drop-low-coverage` for consistency ?
- As some "files" are only produced internally and not written, I simulated them adding a ".tmp" suffix

==> To sum up, I am not very confident about what I wrote here ! (commit afbd0f7ff1a4dc21cc8917f9cdc1f4154a758904)

doc/pipeline.rst
tetedange13: Looking at this line from `batch.py`:
https://github.com/etal/cnvkit/blob/25d91cdb962639c98d041c3b05c7de659c41dc70/cnvlib/batch.py#L210
=> I am not sure if this is really a typo
=> Because conceptually, results produced are "called CNV" so closer to a ".cns" than to a ".cnr", where no CNV calling is made yet (even if this does not match actual file format)
=> Plus this is done late into calling pipeline, so it would probably be distrurbing for a user to have this second ".bintest.cnr" ?

Maybe a solution to improve consistency would be to **actually write a ".cns", but with a dummy "probes" column always at "1"**? (regarding [doc about ".cns"](https://cnvkit.readthedocs.io/en/stable/fileformats.html#segmented-log2-ratios-cns))

etal: The default "probes=1" is a good idea.

The reason this is .cns rather than .cnr is that these are meant to represent copy number calls rather than input data. If `scatter` is extended to allow multiple .cns files as input, then these bin-level calls and the coarse-grained segmentation could be plotted together and would appear as bright colors (a colormap starting with darkorange) rather than as gray raw data points.

doc/pipeline.rst
etal: Adding `--drop-low-coverage` to `segmetrics` is a good idea, and it should match `batch` and the other commands.

`segmetrics` uses a narrow confidence interval for the default filtering approach in `batch` because wide confidence intervals can lead to some pathological behavior -- the end effect is to merge adjacent segments with similar log2 values or a lot of noise in their probes. It's not the same as the p-value alpha. A default of 0.1 might be reasonable but I haven't benchmarked it, so 0.5 is the "safe" approach to only filter out very noisy segments in the final calling stage of the pipeline.

etal (approved):
Very helpful, thanks!

Issue #738, opened 2022-05-05
Hard coded additional GC content filter in fix.py module 


tsivaarumugam:
Hello,

In cnvkit version 0.9.8 and above, there is a hard coded GC content filter at line number 125 in the **"mask_bad_bins"** definition in fix.py module [commit](https://github.com/etal/cnvkit/commit/2e6a40a6664f60cd6743b0c86c917ce4c33c7ab0).

I understand the reason behind the filter and referred the supporting article/publication as well, but this new condition is filters out approximately 10% of our regions of interest.

Is it possible to club this filtering condition with the existing --no-gc flag command line argument and add the hard coded values to params.py module along with preexisting hard code values, so that the filter can be toggled on or off by the user without modifying the code.
 
Attaching the screenshots for reference.

Kindly go through and let me know.

**screenshot of code snippet cnvkits version 0.9.6**
![cnvkits_branch0 9 6_function_mask_bad_bins](https://user-images.githubusercontent.com/92871687/166905662-10187e3a-f47a-47c9-aa87-795f39b7c6a0.png)

 **screenshot of code snippet cnvkits version 0.9.8**
![cnvkits_branch_0 9 8_function_mask_bad_bins](https://user-images.githubusercontent.com/92871687/166905768-377abc56-72e6-4106-9f74-9eb36d61c595.png)

Thanks

tetedange13:
Hi @tsivaarumugam,
<br>

I took a look and I do not think it is possible to connect this hard-coded GC filter you mention, to `--no-gc` param
=> Simply because they do not actionnate the same thing :

* `--no-gc` param toggles whether or not "correction for GC biais" is made (`fix_gc==False` if `fix --no-gc`) :
https://github.com/etal/cnvkit/blob/e29ec7de7bdd03892820a7a712c284c46b580dda/cnvlib/fix.py#L86-L90

* Whereas `mask_bad_bins()` is always run, in order to filter bad bins (*ie* ones having extreme values of GC-content, as shown by your 2nd screenshot) :
https://github.com/etal/cnvkit/blob/e29ec7de7bdd03892820a7a712c284c46b580dda/cnvlib/fix.py#L72

But I can probably submit a PR adding a new `EXTREME_GC_FRACTION` param to `params.py`, as you suggested
=> And you will just have to set it to `1` to disable GC-masking of bad bins
<br>

Hope this helps !
Have a nice day.
Felix.

tsivaarumugam:
Hi Felix,

[...]

PR #753, opened 2022-08-24
Add 'GC fraction for bad bins masking' to PARAMS

tetedange13:
Should adress #738.

Maybe my comment above new `MASKED_GC` in `params.py` could be more explicit about the fact that you set a single value, which is used to perform a symetric filtering ? [eg.: "0.3" filters bin with GC-content "less than 0.3" and "above (1-0.3)=0.7"]

### DIFF
--- a/cnvlib/fix.py
+++ b/cnvlib/fix.py
@@ -122,7 +122,10 @@ def mask_bad_bins(cnarr):
     if 'depth' in cnarr:
         mask |= cnarr['depth'] == 0
     if 'gc' in cnarr:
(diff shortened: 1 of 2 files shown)


etal:
I like this solution. Would you be able to update this PR to specify two thresholds (GC_MIN_FRACTION, GC_MAX_FRACTION) instead of assuming symmetry?

tetedange13:
Done @etal ! :)

Have a nice day !
Felix.

etal:
Awesome, thanks so much!

Issue #606, opened 2021-04-28
Bump networkx dependency to 2.5


auberginekenobi:
Python 3.9 fractions.py is deprecated, making networkx 2.3 raise error attached at bottom. 
Solved in https://github.com/smicallef/spiderfoot/issues/1124 by bumping networkx version to minimum 2.5.
```
Traceback (most recent call last):
  File "/home/ochapman/miniconda3/envs/cnvkit/bin/cnvkit.py", line 4, in <module>
    from cnvlib import commands
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/cnvlib/__init__.py", line 5, in <module>
    from .commands import *
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/cnvlib/commands.py", line 34, in <module>
    from . import (access, antitarget, autobin, batch, bintest, call, core,
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/cnvlib/batch.py", line 8, in <module>
    from . import (access, antitarget, autobin, bintest, call, core, coverage,
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/cnvlib/segmentation/__init__.py", line 17, in <module>
    from . import cbs, flasso, haar, hmm, none
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/cnvlib/segmentation/hmm.py", line 11, in <module>
    import pomegranate as pom
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/pomegranate/__init__.py", line 12, in <module>
    from .parallel import *
  File "pomegranate/parallel.pyx", line 7, in init pomegranate.parallel
  File "pomegranate/hmm.pyx", line 13, in init pomegranate.hmm
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/networkx/__init__.py", line 114, in <module>
    import networkx.generators
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/networkx/generators/__init__.py", line 14, in <module>
    from networkx.generators.intersection import *
  File "/home/ochapman/miniconda3/envs/cnvkit/lib/python3.9/site-packages/networkx/generators/intersection.py", line 13, in <module>
    from networkx.algorithms import bipartite
[...]

PR #771, opened 2022-11-02
Remove no-longer needed networkx pinning for Py3.9

DavidCain:
Newer versions of Pomegranate have explicitly added pinning for
`networkx`. Versions 0.14.5 or later should now automatically pin
`networkx` to a minimum of 2.4

https://github.com/jmschrei/pomegranate/commit/eb88995394690090e808dacf6697b7d665278367

There's no longer a need to do this pinning to ensure Python 3.9
compatibility.

Relates to https://github.com/etal/cnvkit/issues/606

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,10 +21,6 @@ install_requires = [
     'scikit-learn',
     'scipy >= 0.15.0',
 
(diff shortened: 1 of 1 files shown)


etal:
That's great, thanks for the PR!

Issue #784, opened 2023-01-19
Optionally consider PAR1/2 as diploid in male samples due to reference genome masking


rollf:
Hi,

I'd like to address a problem that was already mentioned by @gabeng in this [issue](https://github.com/etal/cnvkit/issues/267). The pseudoautosomal regions in chr Y are masked (by `N`) in hg38 and it also common practice to do so in hg19/GRCh37-based pipelines (see for reference this [GATK blog entry](https://gatk.broadinstitute.org/hc/en-us/articles/360035890951-Human-genome-reference-builds-GRCh38-or-hg38-b37-hg19)). In consequence, the coverage of male samples within the PAR1/2 on chromosome X appears 'doubled'. For example, I have a couple of whole exome samples where one can clearly see this based on the cnn coverage file:

![male01 targetcoverage cnn log2 X0-2500](https://user-images.githubusercontent.com/58295931/213410501-ecc270cf-c947-42a5-ad0e-667352344aef.png)
![male01 targetcoverage cnn depth X0-2500](https://user-images.githubusercontent.com/58295931/213410493-867ba2e7-6594-426c-a865-7a236521c106.png)

The plots show the first 2500 target regions. In contrast, see a female sample:

![female01 targetcoverage cnn depth X0-2500](https://user-images.githubusercontent.com/58295931/213411697-70850d85-0e2c-48d0-8974-582e37f24d43.png)

--- 

Feeding a male-only pool through the usual cnvkit workflow (`reference`, `fix`, `segment`), the previously doubled coverage in PAR1 within the male samples gets lost. The sample has log2 ~ -1 across the whole chromosome.

![male01 cnr master log2 X0-2500](https://user-images.githubusercontent.com/58295931/213410894-2fa14666-8730-49ea-9328-74b816eca588.png)

Furthermore, feeding a mixed pool (5 male, 5 female samples) into the pipeline, the PAR1 region in both samples gets biased. Male samples have log2 ~ -0.5 (i.e. higher than in the male-only pool) and female samples have log2 ~ -0.5 (i.e. lower than in a female-only pool), too (in PAR1 only):

![male01 mixed cnr master log2 X0-2500](https://user-images.githubusercontent.com/58295931/213411490-bbb50bcc-60a1-4685-af7c-0d6e75f970a2.png)
![female01 mixed cnr master log2 X0-2500](https://user-images.githubusercontent.com/58295931/213411533-657b74fe-940f-4c07-97af-df6821e2564d.png)

---

[...]

PR #789, opened 2023-02-20
Consider PAR1/2 on Chr X as autosomal

rollf:
This PR addresses issue [#789](https://github.com/etal/cnvkit/issues/784). It adds the optional feature to treat the PAR1/2 on chromosome X as autosomal, i.e. when requesting autosomal regions these regions are included and on the other hand, when requesting "chromosome X", they are excluded. For chr Y, the PAR1/2 regions are excluded when requesting "chromosome Y" since they cannot be covered if all reads map to PAR1/2 on X. The results have already been shown in the issue mentioned above.

This PR contains the following changes:
* Add new command line argument `--diploid-parx-genome` which can be set to `grch37` or `grch38` in order to treat PAR1/2 GRCh37/38 regions on chromosome X as autosomal.
* This argument is added to all subcommands where necessary and forwarded.
* The method `GenomicArray.autosomes()` has been re-implemented in `CopyNumArray.autosomes()` to be able to transparently include/exclude PAR1/2 on demand.
* Conversely, wherever chromosome X regions have been requested (ex: `(cnarr.chromosome == cnarr._chr_x_label)`), this has been replaced by a call to `CopyNumArray.chr_x_filter()` which includes/excludes PAR1/2 as needed. Same for chromosome Y.
* No differentiation is done based on the sex of the sample.
* This approach has the advantage of having only one place where if/else semantics have to implemented. Whenever the code treats chromosome X, one doesn't have to reason about whether PAR1/2 is include or not.
*  This approach has disadvantages: If enabled, the regions for chrX will be split even for female samples (where it doesn't make sense, as opposed to for male samples [see #784 for an explanation]). Also, one has to rely on the user to write `chrx = self[self.chrx_filter()]` instead of `self[self.chromosome == self.chr_x_label]` (or even `self[self.chromosome == "chrX"]) - but it's in the nature of Python to give full access to object members. 
* Fix a bug in `call.absolute_dataframe()`.
* Add `read_ga` similar to `read_cna()` and use where appropriate (=GenomicArray tests)
* I fixed the version of pomegranate to 0.14.8. See comments.

I will add some further review comments.


cnvlib/call.py
```diff
@@ -148,15 +148,9 @@ def absolute_threshold(cnarr, ploidy, thresholds, is_reference_male):
...
     """Calculate absolute copy number values from segment or bin log2 ratios."""
-    absolutes = np.zeros(len(cnarr), dtype=np.float_)
-    for i, row in enumerate(cnarr):
-        # TODO by_chromosome to reduce number of calls to this
-        ref_copies, expect_copies = _reference_expect_copies(
-            row.chromosome, ploidy, is_sample_female, is_reference_male)
-        absolutes[i] = _log2_ratio_to_absolute(
-            row.log2, ref_copies, expect_copies, purity)
-    return absolutes
+    df = absolute_dataframe(cnarr, ploidy, purity, is_reference_male, is_sample_female)
```
rollf, line 151: `absolute_clonal()` uses a subset of the results from `absolute_dataframe()` so I figured this could be refactored here.

cnvlib/call.py
```diff
@@ -170,18 +164,9 @@ def absolute_pure(cnarr, ploidy, is_reference_male):
 
 def absolute_dataframe(cnarr, ploidy, purity, is_reference_male, is_sample_female):
     """Absolute, expected and reference copy number in a DataFrame."""
-    absolutes = np.zeros(len(cnarr), dtype=np.float_)
-    reference_copies = expect_copies = np.zeros(len(cnarr), dtype=np.int_)
```
rollf, line 174: This is a bug here (at least, I don't think this behavior was intentional). Consider this:

absolute_dataframe_bug

(I.e. any change of either `reference_copies` or `expect_copies` will get reflected in both.)

cnvlib/call.py
```diff
@@ -231,16 +216,18 @@ def _reference_expect_copies(chrom, ploidy, is_sample_female, is_reference_male)
...
-    if chrom in ["chrx", "x"]:
-        ref_copies = (ploidy // 2 if is_reference_male else ploidy)
-        exp_copies = (ploidy if is_sample_female else ploidy // 2)
-    elif chrom in ["chry", "y"]:
-        ref_copies = ploidy // 2
-        exp_copies = (0 if is_sample_female else ploidy // 2)
-    else:
-        ref_copies = exp_copies = ploidy
-    return ref_copies, exp_copies
+    df = cnarr.copy().data
```
rollf, line 219: Direct access of the `data` member. I initially had implemented a `CopyNumArray.get_copied_dataframe()` but realized that at several places throughout the code base `data` is accessed directly. One could replace all those with a dedicated method, though.

skgenome/gary.py
```diff
@@ -205,17 +205,21 @@ def sample_id(self):
...
     # Traversal
 
-    def autosomes(self, also=()):
+    def autosomes(self, also=None):
         """Select chromosomes w/ integer names, ignoring any 'chr' prefixes."""
         is_auto = self.chromosome.str.match(r"(chr)?\d+$", na=False)
         if not is_auto.any():
             # The autosomes, if any, are not named with plain integers
             return self
-        if also:
```
rollf, line 214: This condition doesn't really play nice with pandas. Thus I changed the default value from `()` to `None` and made the check explicit.

test/test_commands.py
```diff
@@ -207,6 +207,77 @@ def test_chrom_means(segments):
...
+        matrix[False][False].append((9, "chrX", 6.93, 2, 1))
+        matrix[False][False].append((10, "chrX", 12.27, 2, 1))
+        matrix[False][False].append((11, "chrX", 4.58, 2, 1))
+        matrix[False][False].append((16, "chrY", 3.13, 1, 1))  # ref_copies = 1 even for female reference
+
+        ret['matrix'] = matrix
+        return ret
+
+    def test_call_absolute_clonal_and_dataframe(self):
+        """Test some helper methods for `call`."""
```
rollf, line 254: I added this test here before fix/refactoring `call.absolute_clonal()`/`absolute_dataframe()` so that I'm sure I'm improving things here.

test/test_genome.py
```diff
@@ -11,15 +11,15 @@
...
+from cnvlib import read, read_ga
 from skgenome import chromsort, rangelabel
 from skgenome import tabio, GenomicArray as GA
 
 
 class GaryTests(unittest.TestCase):
 
     def setUp(self):
-        self.ex_cnr = read('formats/reference-tr.cnn')
+        self.ex_cnr = read_ga('formats/reference-tr.cnn')
```
rollf, line 22: This test is for `GenomicArray` so we should not test a `CopyNumArray` object.

rollf:
@etal Could you have a look into this when time permits?

(I believe the conflicts arose due to new formatting)

test/test_genome.py
```diff
@@ -12,7 +12,7 @@
 import numpy as np
 import pandas as pd
 
-from cnvlib import read
+from cnvlib import read_ga as read
```
rollf, line 15: Make sure to actually test `GenomicArray` not `CopyNumArray`

etal:
Thanks! I'm reviewing this PR now. Sorry for forcing a rebase, but hopefully the codebase is a little easier to work with now.

cnvlib/commands.py
```diff
@@ -86,6 +86,12 @@ def public(fn):
 AP.set_defaults(func=lambda args: AP.print_help())
 AP_subparsers = AP.add_subparsers(help="Sub-commands (use with -h for more info)")
 
+# Shared parameters
+def add_argument_treat_par_on_chrx_as_autosomal_for_genome_build(P):
+    P.add_argument('--treat-par-on-chrx-as-autosomal-for-genome-build',
```
etal, line 91: Should `P` be `parser`? I see an urge to refactor commands.py more substantially this way and collect other reused options into functions like this, but it can be left for the future.

I see help is "todo" here -- how about: "Treat the PAR region of human chromome X as autosomal using its known reference genome coordinates."

cnvlib/export.py
```diff
@@ -27,12 +27,12 @@ def label_with_gene(cnarr):
 
     if not filenames:
         return []
-    first_cnarr = read_cna(filenames[0])
+    first_cnarr = read_cna(filenames[0], None)
```
etal, line 30: It's not obvious what the second argument means at this point; could it be specified as a keyword argument in each case here and below?

cnvlib/metrics.py
```diff
@@ -6,7 +6,7 @@
 from . import descriptives
 
 
-def do_metrics(cnarrs, segments=None, skip_low=False):
+def do_metrics(cnarrs, segments=None, skip_low=False): # todo: make this work in parallel?
```
etal, line 9: Meaning, run metrics on each cnarr in parallel?

cnvlib/reference.py
```diff
@@ -44,14 +45,15 @@ def bed2probes(bed_fname):
     table["gene"] = regions.data["gene"] if "gene" in regions.data else "-"
     table["log2"] = 0.0
     table["spread"] = 0.0
-    return CNA(table, {"sample_id": core.fbase(bed_fname)})
+    return CNA(table, {"sample_id": core.fbase(bed_fname)}) # todo: find all invocations of CNA() and forward param
```
etal, line 48: What does this mean? I don't see a new parameter used here.

test/test_cnvlib.py
```diff
@@ -16,24 +16,42 @@
...
+from cnvlib import cnary, core, fix, smoothing, vary, read
 
 
 class CNATests(unittest.TestCase):
     """Tests for the CopyNumArray class."""
 
     def test_empty(self):
         """Instantiate from an empty file."""
-        cnarr = cnvlib.read("formats/empty")
+        cnarr = cnvlib.read("formats/empty", None)
```
etal, line 27: Could the second argument be handled by a default value of None in the signature of `cnvlib.read` instead?

test/test_commands.py
etal: Excellent, thanks

test/test_genome.py
etal: How about importing `read_ga` with is original name and updating the function calls within the unit tests to `read_ga`?

cnvlib/metrics.py
rollf: Yes, same as for `reference`. Both commands seem ripe for parallelization.

cnvlib/cmdutil.py
```diff
@@ -5,11 +5,20 @@
 from skgenome import tabio
 
 from .cnary import CopyNumArray as CNA
+from skgenome import GenomicArray as GA
 
-
-def read_cna(infile, sample_id=None, meta=None):
+# todo: check all invocations of read/read_cna and make sure the parameter is forwarded.
+def read_cna(infile, treat_par_on_chrx_as_autosomal_for_genome_build, into=CNA, sample_id=None, meta=None):
```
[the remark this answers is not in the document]
rollf: I feel like the decision on PAR is important and I wanted to make sure one is forced to think about it when creating the `CNA`. At least here (with `read_cna()`) I could do this. When invoking the ctor directly (or even cast'ing), I cannot force the developer to take the decision on PAR.

I do understand your concern about breakage for anyone using `cnvlib` as dependency. I think in the current code base, nothing breaks for now. In any case, I'm happy to move the parameter to the end (`into`) as well and thus take a default decision. Just let me know.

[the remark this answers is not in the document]
rollf: As for putting it into `meta`, yes, this is a good idea. This would allow to refactor the code and do the `if-else` here instead of this [line](https://github.com/etal/cnvkit/blob/0a1735753f505e5748d89e56f078cdd70177072c/cnvlib/cnary.py#L32).

cnvlib/export.py
rollf: Yes, I can do this. Depends on whether I make the param optional in the first place (your call).

cnvlib/reference.py
rollf: Maybe the wording here is unfortunate. I wanted to make sure a decision is taken, not only for `read_cna` but also when invoking  the constructor (see my other comment).

etal:
> You haven't said whether you're globally fine with my approach but I assume it's okay then. I will change the code towards adding the `treat_par...` parameter into the `meta` dict.

Yes, I'm globally fine with your approach, and I appreciate your efforts here. My only restriction in this PR is that `read_cna(filename)` and `CNA(table)` should work as expected on a non-human genome like yeast. For human-specific behavior I prefer named keyword arguments over positionals, but you can see I haven't been consistent with that either, so I won't fuss too much about it.

rollf:
Hey @etal,

thanks again for your feedack. I have been working on other topics but I had also tried to make headway here. However, I was really unsatisfied with my implementation. Mainly, because I feel like I am not able ensure consistency (in terms of behavior) across the whole code base (i.e. I can't make sure that the necessary parameter(s) is/are always available when creating/dealing with `CopyNumberArray` objects). I'm currently pursuing a more global approach and will of course get back to you once I have something to show.

rollf:
@etal I worked through the code base again and I'm now done. Can you please carefully review this again? I have updated the initial PR comment as well as added further comments since I have changed my mind and found a better way (IMHO) to implement this optional feature. The tests are currently failing due to dependency issues (see my comment in `core.txt`). I had also used `blackd` but refrained from doing so because I felt like I'm not using the same configuration (some parts of the code were formatted which I didn't touch).

Here are 2 x 3 plots. For the male & the female sample from a mixed pool, I have plotted chr X once based on the current master branch, once based on my branch without the feature enabled and once with the feature enabled. The male bias towards a gain and the female bias towards a loss are fixed (last plot for each sex).

male:
male01
male01
male01

female:
female01
female01
female01

rollf:
Hi @etal,

thanks for the feedback. I renamed the file as you suggested. There are still two tests failing. For me, this looks like a dependency issue (after pinning pomegranate). Let me know how you would like to handle this.

rollf:
Hi @etal,

I will be on a longer summer break for July & August. Do you think we could finish this before end of June? I'm happy to support you if needed but I don't see anything I could do for now.

cnvlib/call.py
```diff
@@ -174,54 +168,46 @@ def absolute_pure(cnarr, ploidy, is_reference_male):
...
+        lambda row: _log2_ratio_to_absolute(
+            row["log2"], row["reference"], row["expect"], purity
+        ),
+        axis=1,
     )
+    return df[["absolute", "expect", "reference"]]
 
 
-def absolute_expect(cnarr, ploidy, is_sample_female):
+def absolute_expect(cnarr, ploidy, diploid_parx_genome, is_sample_female):
```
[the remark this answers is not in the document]
etal: With pandas 2's copy-on-write feature there might not be much memory overhead. Could be worth consolidating code.

etal:
Sure. I'm not too fond of the tests `py3.8-min` and `py3.10-macos`. Ignoring those tests, do you feel this PR is ready to merge?

cnvlib/call.py
```diff
@@ -237,16 +231,22 @@ def _reference_expect_copies(chrom, ploidy, is_sample_female, is_reference_male)
...
+    df["expect"] = np.repeat(ploidy, len(df))
+
+    df.loc[df.chromosome == cnarr.chr_x_label, "reference"] = (
+        ploidy // 2 if is_reference_male else ploidy
+    )
+    df.loc[df.chromosome == cnarr.chr_x_label, "expect"] = (
+        ploidy if is_sample_female else ploidy // 2
+    )
+
+    df.loc[df.chromosome == cnarr._chr_y_label, "reference"] = ploidy // 2
```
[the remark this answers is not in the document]
etal: The better fix here would be to change the parameter name `is_reference_male` to `is_haploid_x_reference` or equivalent. I changed the CLI option name from `--male-reference` to `--haploid-x-reference` in commands.py but it looks like I didn't bother renaming arguments in the deeper function signatures.

The reference profiles come in two flavors: XY (formerly "male" reference, now "haploid x") and XXY (never documented as "female"; it's just the default):
https://cnvkit.readthedocs.io/en/stable/sex.html#reference-sex-chromosome-ploidy

Why does the reference profile have Klinefelter syndrome by default? So that male samples' chrY values can be calculated, while plots still look reasonable (for female samples, copy-neutral X will look the same as the autosomes and there won't be any signal for Y). If reference chrY is 0 then we get divide-by-zero problems.

cnvlib/call.py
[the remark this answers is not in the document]
rollf: I realize that the `purity` parameter would be missing and beside, all the methods already invoke `get_as_dframe_and_set_reference_and_expect_copies()` internally. So I'll keep as is.

cnvlib/call.py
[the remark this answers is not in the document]
rollf: Okay, I replaced occurrences of `reference_male` by `is_haploid_x_reference` (except occurrences of `args.male_reference`) and added a comment.

rollf:
@etal Yes, this is ready for merge IMHO.

etal (approved):

etal:
All done, squashed and merged. Thanks for your contribution!

rollf:
Yeah! :rocket:

eesiribloom:
is there a reason not to implement being able to specify/force sex on input for commands like `cnvkit.py batch` ?
It seems like one of the most common issues users have with cnvkit. I myself have many female samples which are incorrectly assumed to be male.

PR #818, opened 2023-05-22
Fix IndexError for export vcf

pwwang:
I'm not entirely sure why this occurred, but while working with SNP data, I attempted to export the outcomes to a VCF file, and it happened.

It appears that the 'bins.end' or 'bins.start' fields are empty `Series`. Assigning them to 'np.nan' seems to bypass the issue.


etal (approved):
LGTM, thanks for the fix!

Issue #833, opened 2023-07-27
the false negative of segment


qzhqzh:
I find a small deletion **chr2:71600701-71602850** in DYSF. 
but when i run ` cnvkit.py segment -m hmm-germline in.cnr -o out.cns` , the deletion is merged to **chr2:10500-90402011** which is not a real CNV.
I have confirm using QPCR, so i think there is a  false negative result of `cnvkit.py segment`.
How can it output the **chr2:71600701-71602850** deletion in cns file？
Of course， i have run `cnvkit.py segment --drop-low-coverage  -m hmm-germline in.cnr -o out.cns`  too.
 
the cnr lines are:
```
chr2    71570598        71570741        DYSF    344.84615384615387      -0.30249144841605385    0.9813324134255905
chr2    71574197        71574371        DYSF    338.5977011494253       0.04639229891935749     0.9935302308672166
chr2    71589579        71589699        DYSF    419.7083333333333
[cnvkit_data.zip](https://github.com/etal/cnvkit/files/12181884/cnvkit_data.zip)
       -0.06350402781637617    0.9822403060109705
chr2    71590189        71590309        DYSF    418.9583333333333       0.1276053443646271      0.9709310859128705
chr2    71598563        71598745        DYSF    406.84615384615387      -0.04497376934547723    0.9902599453992517
chr2    71600701        71600842        DYSF    195.24113475177305      -0.9822075671662628     0.9753169458242956
chr2    71601453        71601573        DYSF    235.35  -0.8695601481590665     0.9836282601368704
chr2    71602730        71602850        DYSF    198.55  -0.8324236907341246     0.9863254496609705
chr2    71611235        71611355        DYSF    364.2583333333333       0.05401022996463663     0.9818654874605705
chr2    71611464        71611626        DYSF    390.41975308641975      -0.08049284125639751    0.9849266885234083
chr2    71612640        71612806        DYSF    345.578313253012        -0.10463650776802523    0.9668357157464919
chr2    71613311        71613431        DYSF    416.51666666666665      0.08575719564731492     0.9776661461000704
```
the cns line is:
```
[...]

PR #835, opened 2023-08-01
Update hmm.py, make the window paramter activate of segment_hmm function

Zhu-Ying:
make the window paramter activate of segment_hmm function
#833


etal:
Thanks, @Zhu-Ying . Was this change enough to fix the false negative you saw in #833 ?

Zhu-Ying:
> Thanks, @Zhu-Ying . Was this change enough to fix the false negative you saw in #833 ? Did it improve your results overall, and not cause any other false positives or negatives elsewhere?

Yes, if I set the window=2, I can't get more CNVs which is smaller size，and the false negative CNV can be called. the window 
 is larger, the sensitivity is lower. I don't know why, but it works.

etal:
Great, thanks for your contribution!

Issue #822, opened 2023-06-01
Update included references for hg38


marchoeppner:
Hi,

this is probably a minor thing, but the documentation as well as the included references (accessibility etc) are very much focused around hg19. However, at this point we should probably consider hg19 defunct (or at least "bad practice"), given that hg38 came out 10 years ago. It would  hence be nice if CNVkit could be updated throughout (incl. documentation) to use and refer to hg38/GRCh38.

/M

serge2016:
```shell
# GRCh38.d1.vd1)
# https://gdc.cancer.gov/about-data/data-harmonization-and-generation/gdc-reference-files
refGenomeURL="https://api.gdc.cancer.gov/data/254f697d-310d-4d7d-a27b-27fbf767a834" # GRCh38.d1.vd1.fa.tar.gz
refGenomeFN0="GRCh38.d1.vd1.fa.tar.gz"
refGenomeFN="${refGenomeFN0%.tar.gz}"
refGenomeV="${refGenomeFN%.*}"

refGenomeFile="$REFDIR/$refGenomeV/$refGenomeFN"
refGenomeDir="$(dirname "$refGenomeFile")"
refGenomeFaiFile="${refGenomeFile}.fai"
refGenomeDictFile="${refGenomeFile%.*}.dict"

runDir="$(pwd)"
cnvkitRefDir="$runDir/$refGenomeV"
accessFile="$cnvkitRefDir/$GENOMEBUILD.bed"
mkdir -p "$runDir" "$cnvkitRefDir"

if [[ ! -s "$accessFile" ]]; then
    $CNVKIT access $refGenomeFile -o $accessFile
fi
```

etal:
CNVkit bundles hg38/GRCh38 reference assets (e.g. `data/access-10kb.hg38.bed`) and the documentation now leads with hg38.

Beyond the original hg38 request, the broader need — running CNVkit on other reference genomes — is now supported: `cnv_gene_info.py` (PR #1110) builds the RNA gene-resource table for any genome, and `cnvkit.py access <genome.fa>` computes the DNA accessible-regions file for any organism. Both are documented in the "Building reference assets for other genomes" section of `doc/rna.rst` (ships in the next release after v0.9.13).

Closing as addressed; please reopen if a specific hg38 or other-genome gap remains.

PR #837, opened 2023-08-09
hg38 refFlat and accessibility data

berguner:
Should help with  #822


etal (approved):

etal:
Thanks! Did these sources come from UCSC?

berguner:
Yes, I generated the refFLat file using UCSC table browser.

Issue #859, opened 2023-11-30
Smoothing window fraction fix results in noisier data in targeted sequencing panel?


kkchau:
We're currently evaluating an update from version 0.9.9 to 0.9.10. However, we're seeing some increased noise among the ratios and resulting in an increased rate of false positives. Specifically, looking at the IQRs:

![image](https://github.com/etal/cnvkit/assets/13711229/4d710176-eddc-41d7-aff5-d5ce8d33cf8c)

We've traced it back to this [commit](https://github.com/etal/cnvkit/commit/3b9c3ba77c11759af97b8ddf9e1dd13b1ce92580); I noticed that the window fraction now defaults to a limit of 0.01 (from the flat 0.1). Could this be under-correcting the log2 ratios during the `fix` step?

```
$ wc -l sample.target.coverage.cnn
43719 sample.target.coverage.cnn

$ wc -l sample.antitarget.coverage.cnn
3143 sample.antitarget.coverage.cnn

$ cnvkit.py fix --no-gc sample.target.coverage.cnn sample.antitarget.coverage.cnn sample.reference.cnn -o sample.cnr -i sample
```

etal:
It could be under-correcting for GC and repeats overall. My aim in that commit was to improve performance on bins with extreme GC, where e.g. Picard showed there was still a correlation between GC content and coverage depth, by using a smaller window where there were enough data points. You have better benchmarking datasets on hand than I do at this point, so if you're seeing a deterioration, then I hope your PR gets you back to where you need to be.

I'm also curious if you find the original 10% was optimal or another window size is even better on today's data.

PR #860, opened 2023-12-05
Allow manual tuning of smoothing window fraction

kkchau:
Adds a parameter to allow for manual tuning of the smoothing window fraction. If not supplied, will fall back to the smoothing fraction function described in https://github.com/etal/cnvkit/commit/3b9c3ba77c11759af97b8ddf9e1dd13b1ce92580.

Addresses https://github.com/etal/cnvkit/issues/859


etal:
LGTM, thanks for the fix!

PR #870, opened 2024-03-01
Use correct PAR1/2 start positions.

rollf:
The start position was wrong because it was assuming a one-based coordinate system. However, these paremters are used in BED-context where coordinates are assumed zero-based. Since BED is right-open (i.e. the end coordinate is *not* within the interval), the end positions of PAR1/2 were actually correct.

For reference, here are some pictures that depict the difference between BED coordinates and "genomic" coordinates:

image-20240301-092030
(From [NCBI](https://www.ncbi.nlm.nih.gov/grc/human))

image-20240301-091943
(`N` here are the telomeres. Chromosome X's first non-masked position (`C`, at 60,001) is the first position of PAR1)

image-20240301-092451

NCBI’s “2,649,520” (PAR1 end) is 2,649,519 in zero-based coordinates. However, BED’s end coordinate is not included (see above the screenshot: Although x-par1.bed contains 60,000 and 60,010 there are 10 (not 11) bases extracted, i.e. base 60,000 to base 60,009). Thus, using 2,649,519 as the end coordinate within a BED file would mean, the last base of PAR1 (“2,649,520” in NCBI’s coordinate system) would be omitted. Thus, we need to increment by one. The final BED numbers would be start=60,000;end=2,649,520 which equals NCBI’s start=60,001 and stop=2,649,520. Hence, this PR only decreases all start positions by one.


etal:
Makes sense to me, thanks!

PR #910, opened 2024-09-04
Use pomegranate 1.x API in HMM segmentation mode

etal:


rollf:
Ah, cool @etal Thank you for addressing this!

mr-c:
Cool, it would be nice to upgrade as pomegranate < 1.0 is having problems with numpy 2.x in Debian

apraga:
Waiting for https://github.com/jmschrei/pomegranate/issues/1130 for advice. In the meantime, if you have an idea about how to lift `fit()` in HMM for the new version, @etal, that would be great :)

etal:
The deed is done. I apologize for the bot spam.

@mr-c -- I've split CNVkit into two pip/conda packages, where the only difference is whether the `pomegranate` and `pytorch` dependencies are required; the Python codebase itself is the same either way. For Debian, assuming pytorch and pomegranate 1.0+ are already available as a packages, I think I'd recommend including the HMM dependencies and ignoring the split.

Without the pomegranate dependency (default behavior with `pip install cnvkit` going forward), if you run segmentation with the HMM method (`-m hmm`), it will fail and report the missing dependency. With the dependency (`pip install cnvkit[hmm]`), it should all work as you'd expect.

Issue #898, opened 2024-07-18
Installation problems


EfraMP:
Hi, 

I'm a MacOS. facing some problems.

First, I tried to install cnvkit via conda, but I was unable and got the next output:

```
Channels:
 - conda-forge
 - bioconda
 - defaults
Platform: osx-arm64
Collecting package metadata (repodata.json): done
Solving environment: failed

LibMambaUnsatisfiableError: Encountered problems while solving:
  - nothing provides bioconductor-dnacopy needed by cnvkit-0.9.10-pyhdfd78af_0

Could not solve for environment specs
The following package could not be installed
└─ cnvkit is not installable because it requires
   └─ bioconductor-dnacopy, which does not exist (perhaps a missing channel).
```

Them I attempted to use pip, but got the next output:

```
Collecting cnvkit
  Downloading CNVkit-0.9.11.tar.gz (176 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 176.3/176.3 kB 1.9 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: biopython>=1.80 in /Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages (from cnvkit) (1.81)
Requirement already satisfied: matplotlib>=3.5.2 in /Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages (from cnvkit) (3.8.1)
Requirement already satisfied: numpy>=1.24.2 in /Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages (from cnvkit) (1.26.1)
Requirement already satisfied: pandas>=1.5.3 in /Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages (from cnvkit) (2.1.2)
Collecting pomegranate<=0.14.9,>=0.14.8 (from cnvkit)
  Downloading pomegranate-0.14.9.tar.gz (4.7 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 4.7/4.7 MB 5.5 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
[...]

Issue #890, opened 2024-06-22
Installation error


mikewojnowicz:
Hi, I'm running Python 3.11.5.   I am running a 2023 Macbook Pro; hence, I am using the arm64 version of Python (for the M1 or M2 chips), and not the  x86_64 version of Python (for an Intel chip).  The command `pip install cnvkit` gave the following error message below.  Do you have any advice? Thank you.

```
⇒  pip install cnvkit
Collecting cnvkit
  Using cached CNVkit-0.9.11.tar.gz (176 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting biopython>=1.80 (from cnvkit)
  Using cached biopython-1.83-cp311-cp311-macosx_11_0_arm64.whl.metadata (13 kB)
Requirement already satisfied: matplotlib>=3.5.2 in ./deactivate/lib/python3.11/site-packages (from cnvkit) (3.9.0)
Requirement already satisfied: numpy>=1.24.2 in ./deactivate/lib/python3.11/site-packages (from cnvkit) (1.26.4)
Requirement already satisfied: pandas>=1.5.3 in ./deactivate/lib/python3.11/site-packages (from cnvkit) (2.2.2)
Collecting pomegranate<=0.14.9,>=0.14.8 (from cnvkit)
  Using cached pomegranate-0.14.9.tar.gz (4.7 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [75 lines of output]
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
      
      cpdef disable_gpu():
          global GPU
          GPU = False
      
      cdef ndarray_wrap_cpointer(void* data, numpy.npy_intp n):
          ^
      ------------------------------------------------------------
      
      pomegranate/utils.pyx:120:5: Function signature does not match previous declaration
      Compiling pomegranate/BayesClassifier.pyx because it depends on /private/var/folders/vg/q45k1lj96hzd0qyfkwstbqbr0000gq/T/pip-build-env-fgw3yhvh/overlay/lib/python3.11/site-packages/Cython/Includes/libc/string.pxd.
[...]

PR #911, opened 2024-09-04
Raise max pomegranate dependency version from <=0.14.9 to <1.0.0

etal:
Via issues #898 and #890.


etal:
The "min" tests fail because pomegranate 0.14.8 won't compile successfully when built this way, but a separately precompiled installation will (I think) pass the tests. When installing with conda, pomegranate 0.15.0 is the only version that works.

PR #912, opened 2024-09-11
Cast min_mapq to str instead of bytes

rach-kennedy:
Hi, I've been evaluating cnvkit for our data and overall liking it - results are looking good! However I've been noticing that the `--min-mapq` parameter doesn't seem to be propagating through correctly, and believe that this one-line change would be a fix.

To replicate the issue using data already present in this repo, you can pull out this single read with MAPQ=33 into a mini bam:
```
> samtools view -h test/formats/na12878-chrM-Y-trunc.bam | grep -e HWI-ST1124:106:C15APACXX:1:1107:19114:135147 -e ^@ | samtools view -hb > single_read_qual_33.bam && samtools index single_read_qual_33.bam
> echo -e "chrM\t7000\t8000" > single_target.bed
```

And then run `cnvkit.py coverage` with two different thresholds, which yield identical .cnn files despite falling above and below the MAPQ of this read:

```
> cnvkit.py coverage -q 0 single_read_qual_33.bam single_target.bed -o cov_qual_0.cnn
> cnvkit.py coverage -q 60 single_read_qual_33.bam single_target.bed -o cov_qual_60.cnn
```

And poking around more in the python, here's how using `str` instead of `bytes` is fixing the behavior:

```
>>> import pysam
>>> bed_fname = "single_target.bed"
>>> bam_fname = "single_read_qual_33.bam"
>>> pysam.bedcov(*[bed_fname, bam_fname, "-Q", bytes(60)])
'chrM\t7000\t8000\t200\n'
>>> pysam.bedcov(*[bed_fname, bam_fname, "-Q", str(60)])
'chrM\t7000\t8000\t0\n'
```

Thank you, and let me know if I've misunderstood something!


etal:
Thanks for the fix! This makes sense.

etal (approved):

PR #934, opened 2024-12-20
Remove numpy `asfarray` and `float_` usage to support v2.0 deprecation

suhas-r:
`np.asfarray`+`np.float_` were removed in Numpy-2.0.0 ([release notes](https://numpy.org/devdocs/release/2.0.0-notes.html)), in favor of using `np.asarray` / `np.float64`, respectively. [np.asarray](https://numpy.org/doc/1.24/reference/generated/numpy.asarray.html#numpy-asarray) + [np.float64](https://numpy.org/doc/1.24/reference/arrays.scalars.html#numpy.float64) are available i existing `numpy>=1.24.2` versions currently required by the cnvkit [conda env](https://github.com/etal/cnvkit/blob/master/conda-env.yml#L18). 

cnvkit otherwise functions properly with`numpy-2.*` and `python-3.12` after updating these methods.

### DIFF
--- a/cnvlib/bintest.py
+++ b/cnvlib/bintest.py
@@ -88,7 +88,7 @@ def z_prob(cnarr):
 
 def p_adjust_bh(p):
     """Benjamini-Hochberg p-value correction for multiple hypothesis testing."""
-    p = np.asfarray(p)
+    p = np.asarray(p, dtype=float)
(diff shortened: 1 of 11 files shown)


etal (approved):

etal:
Fantastic, thanks for the update!

Issue #781, opened 2022-12-27
batch creates FASTA's bed in working dir not in output-dir


solo7773:
version 0.9.9

cnvkit.py batch --normal $normal
--fasta $fasta
--output-dir $outdir
-m wgs

I use the above command to generate the `reference.cnn`. The command runs in a computing cluster via qsub. The command runs as expected, excepting a file named `$fasta.bed` is generated in the working dir. I checked and found it's the output of the `access` subcommand. Could you fix this bug, although it's not fatal?

etal:
Fixed in v0.9.13 (commit 0003b20, "generate target_bed in output_dir if given").

For the record: in a WGS batch run with `--fasta` but no `--access`/`--targets`, CNVkit runs `access` on the fly and writes its BED to disk. In 0.9.9 that write used only the FASTA basename, so it landed in the working directory rather than `--output-dir`:

```python
target_bed = os.path.splitext(os.path.basename(fasta))[0] + ".bed"
tabio.write(access_arr, target_bed, "bed3")   # bare basename -> CWD
```

Current releases join `--output-dir` before writing:

```python
target_bed = os.path.splitext(os.path.basename(fasta))[0] + ".bed"
if output_dir:
    target_bed = os.path.join(output_dir, target_bed)
tabio.write(access_arr, target_bed, "bed3")
```

Upgrading to v0.9.13 or later resolves this. Thanks for the report.

PR #940, opened 2025-01-22
cnvkit batch - generate target_bed in output_dir if given

pontushojer:
Hi!

Small edit here.

I noticed that the target BED generated by `cnvkit.py batch` was created in the working directory, NOT the specified output directory. This PR add some logic to fix this.

For context I am running `cnvkit batch` in WGS mode using only a FASTA while specifying an output directory, this is the command:

```
cnvkit.py batch path/to/my.cram -n -m wgs -f path/to/genome.fa -d output
```

This generates a target BED based on the FASTA, but despite giving an output directory this BED (`genome.bed` in this case) was generated in the current directory.


pontushojer:
Relates to: https://github.com/etal/cnvkit/issues/781

etal (approved):

etal:
Thanks!

PR #956, opened 2025-06-19
fix: known extensions mapping in sniff_region_format()

dlaehnemann:
This change avoids always mismatching, because previously the `.` from the extension was already removed with `ext = ext.lstrip('.')` two lines above. Thus, the `ext[1:]` (which was probably put there to also remove the dot) removes the first letter of the actual file extension and the matching with the keys of `format_patterns` just always returns `False`. This change should fix this.


etal:
Thanks for the fix!

Issue #971, opened 2025-10-23
cnvkit WES tumor only fails to generate scatter plots


asmariyaz23:
Hello,

I am trying to run cnvkit in the following steps, for context I have tumor only samples from WES dataset:

```
cnvkit.py \
    antitarget \
    intervals_sorted.bed \
    --output intervals.antitarget.bed

cnvkit.py \
    reference \
    --fasta Homo_sapiens_assembly38.fasta \
    --targets intervals_sorted.bed \
    --antitargets intervals.antitarget.bed \
    --output cnvkit.reference.cnn

cnvkit.py \
    batch \
    OV2295_P61.recal.bam \
    --reference cnvkit.reference.cnn \
    --processes 2 \
    --method hybrid --diagram --scatter
```

This only generates the antitarget and target.cnn files (scatter plots, cnr etc files are not generated). Following is the log from BATCH command:

```
Wrote ./cnvkit.reference.target-tmp.bed with 242421 regions
Wrote ./cnvkit.reference.antitarget-tmp.bed with 42709 regions
Running 1 samples in 2 processes (that's 2 processes per bam)
Running the CNVkit pipeline on work/cd/2c62e2ef6ca2dcfa7e8ef4a91381b3/OV2085_P64.recal.bam ...
Processing reads in OV2085_P64.recal.bam
/scratch/asmab/cnvkit/lib/python3.11/site-packages/cnvlib/coverage.py:198: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value '[119.39108911 304.50406504 350.65671642 ...   0.75         0.93969849
   0.        ]' has dtype incompatible with int64, please explicitly cast to a compatible dtype first.
  table.loc[ok_idx, "depth"] = table.loc[ok_idx, "basecount"] / spans[ok_idx]
Time: 135.528 seconds (391138 reads/sec, 1789 bins/sec)
Summary: #bins=242421, #reads=53010139, mean=218.6698, min=0.0, max=70315.11
Percent reads in regions: 68.456 (of 77436567 mapped)
Wrote ./OV2085_P64.targetcoverage.cnn with 242421 regions
Processing reads in OV2085_P64.recal.bam
[...]

Issue #951, opened 2025-04-26
Genemetrics Indexing issue


vanshikabidhan:
Hi @etal! Thanks for developing and maintaining the tool. I am trying to call CNVs at gene-level resolution in my amplicon data. Note that I am not analysing tumour samples but rather some other disease, so I am taking non-patient samples as "control" here. Furthermore, my data has nearly 200 amplicons, with an average length of 230bp (50x coverage). 

These are the steps that I follow:
1. Create a dummy reference : cnvkit.py reference -o flatRef.cnn -f hg19.fa -t targets.bed
2. Create a "real" reference using the "control" samples: 
     cnvkit.py batch -m amplicon -r flatRef.cnn dir/*.bam -p 8
     cnvkit.py reference dir/*coverage.cnn -f hg19.fa -o controlsRef.cnn
3. Re-run on patient samples using controlsRef.cnn: cnvkit.py batch -m amplicon -r controlsRef.cnn samples/*.bam -p 8

While the workflow above performs segmentation and reports the results at chromosomal level, I also used genemetrics with just the cnr files, to be able to call CNVs across the genes:
"" for file in dir/*.cnr; do
    prefix=$(basename "$file" .cnr)
    cnvkit.py genemetrics "$file" -o "${prefix}.tsv" 
done ""

However, I observed that the function takes gene intervals incorrectly. For example:

A few amplicons in my panel look like this:
chr4	89858038	898xxxxx	gene1
chr4	90647605	906xxxxx	gene2
chr4	90650238	906xxxxx	gene2
chr4	907xxxxx	907xxxxx	gene2
chr4	907xxxxx	907xxxxx	gene2
chr4	907xxxxx	907xxxxx	gene2
chr4	106153834	106154062	gene3
chr4	106154101	106154312	gene3

However, Genemetrics defines gene2 with the following coordinates:
gene2	chr4	90650238	106154312 (thus picking up the coordinate from gene3)

At the same time, I also noticed that it never picks up the first starting position of any gene. This could be attributed to some failed amplicons (whereby the bins fail filters), however this does not seem to be the case. Based on the issue #773, I suspected incorrect indexing in cnary.py file, and attempted to modify the 'by_gene' function. However, a lot of other function dependencies exist. Could you please, firstly let me know if there is something that I am doing wrong, or if you have looked in this issue? 

Thanks a lot, 
V.

Issue #773, opened 2022-11-09
genemetrics result different between  0.9.8  and 0.9.9 


pmtmarine:
cnary.py  
```
def by_gene(self, ignore=params.IGNORE_GENE_NAMES):
   ignore += params.ANTITARGET_ALIASES
   start_idx = end_idx = None (only del in 0.9.9 ? Why?)

```

something wrong in genemetrics result 
```
$ python3 cnvkit.py genemetrics sample.cnr >sample.final_gene.cnr

$ grep -w MYC sample.final_gene.cnr
chr8	128747718	128908053	MYC	6.36412	165	12089.4	9	8.6922

$ grep -w MYC sample.cnr -C 2
chr8	128429691	128588454	Antitarget	160.124	6.90919	0.970921
chr8	128588454	128747218	Antitarget	239.995	7.32225	0.986018
chr8	128747718	128748063	MYC	10755.5	6.03366	0.953326
chr8	128748291	128748580	MYC	13712.4	6.51247	0.956057
chr8	128748580	128748869	MYC	15145.9	6.23034	0.96773
chr8	128750493	128750750	MYC	14552	6.44618	0.965993
chr8	128750750	128751007	MYC	16226.7	6.50782	0.967813
chr8	128751007	128751265	MYC	14297.6	6.42488	0.966846
chr8	128752641	128752922	MYC	12428.5	5.86496	0.971669
chr8	128752922	128753204	MYC	11518.5	6.05504	0.969818
chr8	128753704	128908053	Antitarget	237.28	7.19697	0.972948
chr8	128908053	129062402	Antitarget	293.876	7.4257	0.981591
```
The end  num  should be **128753204**  not  **_128908053_**	!!!!!!

pmtmarine:
I found that the reasons for this difference were `loc` and `iloc` in  cnary.py  def by_gene

etal:
Thanks for your report -- lost but not forgotten. I've finally fixed this bug with #979.

PR #979, opened 2026-01-31
Fix interval indexing bug and batch error suppression

etal:
Via: #971, #951, #773

Fixes:

1. Batch Command Exception Handling (cnvlib/commands.py:211-234)
- Problem: Exceptions in parallel worker processes were silently swallowed
- Fix: Collect and wait for all futures, calling .result() to propagate exceptions

2. SerialPool Exception Handling (cnvlib/parallel.py:24-51)
- Problem: SerialFuture didn't capture exceptions
- Fix: Store exceptions and re-raise them when .result() is called

3. Duplicate Coordinates in by_gene() (cnvlib/cnary.py:159-185)
- Problem: iloc/loc confusion caused overlapping regions
- Fix: Convert index labels to positions and use .iloc[] for slicing

### DIFF
--- a/test/test_commands.py
+++ b/test/test_commands.py
@@ -856,7 +856,7 @@ class CommandTests(unittest.TestCase):
                     "depth": [100.0, 200.0, 200.0, 150.0],
                 }
             )
-            target_cnarr = cnvlib.CopyNumArray(target_data, {"sample_id": "test_sample"})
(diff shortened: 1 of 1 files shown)


etal:
Also looking at these issues that suggest a deeper interval-arithmetic bug:
https://github.com/etal/cnvkit/issues/951
https://github.com/etal/cnvkit/issues/773

Issue #616, opened 2021-05-09
Improve documentation on `cnvkit call` and the use of VCFs


tskir:
Currently, some aspects of how VCF samples are loaded and processed are a bit obscure in the documentation. This especially applies to things like BAF rescaling and LOH detection; see for example https://github.com/etal/cnvkit/issues/601#issuecomment-835861846.

It seems like a good idea to restructure and clarify the parts related to VCF, how the samples are treated, how the maths of BAF rescaling and LOH detection works, etc.

Issue #601, opened 2021-04-26
negative BAF values estimating LOH


john-alexander:
I'm trying to estimate LOH in cnvkit using purity, ploidy estimates from sequenza. 

I use cnvkit call option with its default thresholds, 

    conda activate **cnvkit0.9.7**
    cnvkit.py call $tumour_id.recalibrated.cns -y -v $mutect_tumour_dir/$tumour_id/$tumour_id.passed.vcf.gz --purity $purity --ploidy $ploidy -o $tumour_id.call.cns

The output gives me -ve BAF values which im not sure how to interpret

      chromosome    start      end        gene                                   			log2       baf 	cn 	cn1 cn2   depth probes  weight
    chr16 66669013 69943499 CMTM4,DYNC1LI2,DYNC1LI2,LOC106699570,LOC106699570,LOC106699570,TERB1  -0.871623 -0.211176  1   1   0 71.7805   2818 2418.18

To confirm are LOH regions where cn1!=cn2 and BAF=0. Also  what about estimating copy number netural LOH.

Thank you.

tskir:
Welp, that doesn't seem good.

To debug this further, I'm afraid I'm going to need a minimal reproducible example with the files & the exact parameters you're using. If these are private data, feel free to only share the relevant chunks and/or to randomise some values, as long as the bug (the negative BAFs) can still be reproduced. Do you think you could provide something like that?

john-alexander:
I'm sending private files via zendto (ktsukanov [at] ebi.ac.uk) 

Command used: 

```
conda activate cnvkit0.9.7
cnvkit.py call tumour_AB.recalibrated.cns -y -m threshold -t=-1,-0.3,0.3,1 -v tumour_AB.passed.vcf.gz --purity 0.37 --ploidy 2 -o tumour_AB.call.cns

```
Thank you very much.

tskir:
@john-alexander Thank you, files received! I'll now look into that more closely

john-alexander:
@tskir thank you for looking into this -wondering if there's a fix/update on this soon.
Best,
J

tskir:
Hi @john-alexander, thank you for waiting. I was now able to take a look into this. As far as I can see, the problem has two sides to it.

First, the `cnvkit.py call` command you are using does not specify the tumour and normal sample IDs (`--sample-id`, `--normal-id`). Because of this, only the first sample is used and the whole VCF is treated as unpaired tumour sample with no normal.

[...]

PR #1047, opened 2026-04-29
Issues 601, 616: Avoid negative BAF values, update VCF docs

etal:
Addresses #601 and #616 --

Ensure negative BAF values aren't produced, even from VCFs that break assumptions/expectations.

Update and clarify documentation around VCF handling.


etal:
@claude review

PR #1048, opened 2026-05-06
Relax pyparsing dependency

SkohTV:
https://github.com/pyparsing/pyparsing `3.3.2` is out, and cnvkit still build without issue when upgrading
Is there a reason why it is capped at `3.3.0` ?


etal:
Let me take a look. Here's original commit that imposed the version cap: https://github.com/etal/cnvkit/commit/7b3025c213523c4808a3f09b79f42ffe8853233a 

There was probably a dependency conflict; pyparsing isn't used directly in the codebase, so this would have been to deal with some other circle of dependency hell, most likely.

etal:
Here's the scoop: pyparsing did a PEP8 migration and changed camelCase method names to underscore_style, leaving the old names as shims which matplotlib depended on. But pyparsing 3.3.0 started emitting warnings for the old camel-case calls, and matplotlib 3.9.0 hadn't finished its own PEP8 migration.
- https://github.com/matplotlib/matplotlib/issues/29722

Evidence:

- https://github.com/matplotlib/matplotlib/issues/30617
- https://github.com/matplotlib/matplotlib/issues/25204

Meanwhile, CNVkit treats most library warnings as errors to fight bit rot, so the pyparsing < 3.3.0 version cap was required for the plotting code to function at all with matplotlib 3.9.0.

Fix: raise the matplotlib minimum version to where it switches to pyparsing's new names:
- https://github.com/matplotlib/matplotlib/pull/29745

etal:
Fixed in #1050 -- thanks for pointing out the issue

PR #1062, opened 2026-05-23
Security: Command Injection via R Script Template String Formatting

tomaioo:
## Summary

Security: Command Injection via R Script Template String Formatting

## Problem

**Severity**: `High` | **File**: `cnvlib/segmentation/cbs.py:L1`

In cnvlib/segmentation/cbs.py and cnvlib/segmentation/flasso.py, R script templates use Python string formatting with %(probes_fname)s, %(sample_id)s, and other values that originate from user-controlled input (file paths, sample IDs). These values are interpolated into R scripts that are later executed via subprocess. If an attacker can control these values (e.g., through crafted file names or sample IDs), they could inject malicious R code. While the templates use double quotes for R strings, the lack of proper escaping of the interpolated values creates a command injection risk when the R script is executed.

## Solution

Use proper shell escaping for all interpolated values, or better yet, pass values as command-line arguments to the R script rather than interpolating them into the script template. Consider using subprocess with a list of arguments instead of string formatting for command execution.

## Changes

- `cnvlib/segmentation/cbs.py` (modified)

### DIFF
--- a/cnvlib/segmentation/cbs.py
+++ b/cnvlib/segmentation/cbs.py
@@ -7,8 +7,14 @@ CBS_RSCRIPT = """\
 
 library('DNAcopy')
 
+args <- commandArgs(trailingOnly=TRUE)
+probes_fname <- args[1]
+sample_id <- args[2]
+threshold <- as.numeric(args[3])
+smooth_cbs <- as.logical(args[4])
+
 write("Loading probe coverages into a data frame", stderr())
-tbl = read.delim("%(probes_fname)s")
+tbl = read.delim(probes_fname)
 # Filter the original .cnr to valid rows (usually they're all valid)
 if (!is.null(tbl$weight) && (sum(!is.na(tbl$weight)) > 0)) {
     # Drop any null or 0-weight bins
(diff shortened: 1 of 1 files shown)


etal:
@claude Please review -- you two seem to have something in common.

etal (changes requested):
Tests are failing

tomaioo:
Thanks for the ping, @etal. You're right that we share a name—and hopefully a careful eye for code review. Looking at the PR, passing values as command-line arguments via `subprocess.run([...])` rather than template interpolation would indeed eliminate the injection vector entirely. I'd also suggest validating that `shlex.quote()` or similar isn't just being applied to the R script string itself, which wouldn't help; the values need to be passed out-of-band from the script body.

tomaioo:
Thanks for the ping, @etal. You're right that we share a name—and hopefully a careful eye for code review. Looking at the PR, passing values as command-line arguments via `subprocess.run([...])` rather than template interpolation would indeed eliminate the injection vector entirely. I'd also suggest validating that `shlex.quote()` or similar isn't just being applied to the R script string itself, which wouldn't help; the values need to be passed out-of-band from the script body.

Issue #892, opened 2024-06-26
Getting BAF and allelic imbalance in export vcf


GACGAMA:
Hello everyone!

I've called my segments by using both a VCF with tagged somatic/germline variants, exported to theta2 and and imported back to CNVkit.

Now I need to export the final called CNS file to VCF format to annotate with AnnotSV.
When exporting to VCF, all allelic imbalance and BAF information is lost! Now I don`t know which variants could be LoH

I used the following commands to export:
```

#call after importing from theta
cnvkit.py call /scratch4/nsobrei2/ggama1/somatic_SVs/cnvkit/exomes_segments_test/1_13_TUMOR-1.cns -v /scratch4/nsobrei2/ggama1/somatic_SVs/cnvkit/vcfs/germline/somatic_removed/1_13_TUMOR_reordered.vcf.gz -i 1_13_TUMOR -n GERMLINE -o /scratch4/nsobrei2/ggama1/somatic_SVs/cnvkit/exomes_segments_test/1_13_TUMOR_call_theta.cns


#export as vcf


cnvkit.py export vcf /scratch4/nsobrei2/ggama1/somatic_SVs/cnvkit/exomes_segments_test/1_13_TUMOR_call_theta.cns -i "1_13_TUMOR" -o /scratch4/nsobrei2/ggama1/somatic_SVs/cnvkit/exomes_segments_test/1_13_TUMOR.cnv.vcf
```

etal:
Allele-specific copy number and B-allele frequency are now emitted by `export vcf`, as of commit 26f7499 (PR #1099) on the current master branch (ships in the next release after v0.9.13).

When the input `.cns` carries allele-specific columns — produced by `cnvkit.py call -v variants.vcf` — `export vcf` now adds:

- **CN1 / CN2** (FORMAT): major / minor allele copy number, following the ASCAT/Battenberg/PURPLE convention, with CN1 + CN2 = CN.
- **BAF / BAFN** (INFO): per-segment median mirrored B-allele frequency and the number of supporting heterozygous SNPs.
- An **LOH** flag (INFO) for copy-neutral loss of heterozygosity (total copy number matches the expected ploidy but one parental allele is entirely lost), emitted as `SVTYPE=CNV` rather than being silently dropped.

A `.cns` produced without `-v` (no allele-specific columns) yields the previous VCF shape unchanged, so existing downstream parsers are unaffected. This should give AnnotSV the LOH / allelic-imbalance evidence you needed.

[...]

PR #1100, opened 2026-05-29
scatter: surface LOH and somatic SNV evidence as colored overlays (#290)

### LINKED ISSUE #290: Missing points in VAF plot
Maybe I am getting this all wrong; this could very well be a feature and not a bug. I have created a plot using the `scatter` command:

`cnvkit.py scatter test.cnr -s test.cns -v test.vcf.gz -c chr20:19569338-21565309 -o plot.pdf`

plot

There appears to be a loss (and hence LOH) in part of the RALGAPA2 gene. In fact, there is a homozygote somatic variant in the area not segmented in the VAF plot supporting the hypothesis:

```
$ bedtools intersect -a test.vcf.gz -b <(echo -e "chr20\t20500000\t20750000")
chr20	20572016	rs55967804	A	C	33954.93	PASS	AC=2;AF=1.00;AN=2;BaseQRankSum=2.00;ClippingRankSum=0.107;DB;DP=25;ExcessHet=0.2098;FS=1.244;InbreedingCoeff=0.0795;MQ=54.25;MQRankSum=-2.300e-02;POSITIVE_TRAIN_SITE;QD=12.85;ReadPosRankSum=0.578;SOR=0.796;VQSLOD=3.90;culprit=DP	GT:AD:DP:GQ:PL	1/1:1,24:25:38:898,38,0
```

However, there is no corresponding dot in the VAF plot. I have tried fiddling around with the code and extended the y-axis, but there is still no dot. Is that the expected behavior?

I have also tried `--min-variant-depth 1`, but there is still no dot.

etal:
## Summary

- Adds `cnvkit.py scatter --show-snvs {het,with-loh,with-somatic,all}` to overlay tumor-homozygous (LOH-evidence) and somatic SNV loci in the VAF panel with distinct colors and markers, with an external legend identifying each kind. The default `het` mode is unchanged.
- The segment-mean VAF trend is always computed from the het subset only, so the BAF math that drives clinical segment-overlay interpretation is unchanged.
- Closes #290 (open since 2017): "plotting somatic or tumor-homozygous variants in another color."

Visualization-only feature. No change to `.cnr`, `.cns`, `.cnn`, SEG, or VCF outputs.

## Design notes

- **LOH definition.** When matched normal genotypes are available, LOH evidence = normal-heterozygous AND tumor-homozygous, excluding any somatic call. In tumor-only flows the selector falls back to tumor-homozygous (with the documented caveat that true germline-homozygous positions are indistinguishable from LOH-induced homozygosity).
- **Marker style.** Wong colorblind-safe palette: reddish-purple `x` for LOH (typically rare and at VAF extremes; needed visual pop), blue `+` for somatic (typically widespread at low VAF). Distinct shapes so the overlays read in grayscale prints.
- **API shape.** New `cmdutil.load_snp_subsets(...)` returns `(het, loh_or_None, somatic_or_None)`; the het result is identical to `load_het_snps`, so the 8 existing `load_het_snps` callers are unchanged. The Mutect2 all-zero-normal `zygosity_freq` workaround is factored into `_resolve_zygosity_freq` and shared between the two loaders.
- **Plumbing.** Six functions in `cnvlib/scatter.py` thread optional `loh_variants` / `somatic_variants` parameters from `do_scatter` down to the per-chromosome plot calls. `cnvlib/plots.py::update_binwise_positions` accepts optional `extra_variants` and now always returns a 4-tuple.

## Code review

Three parallel `feature-dev:code-reviewer` agents reviewed before push (simplicity/DRY, bugs/correctness, project conventions). Findings addressed:

- `update_binwise_positions` reshaped to always return a 4-tuple (no variable-arity return).
- `_resolve_zygosity_freq` helper extracted to deduplicate the Mutect2 workaround between `load_het_snps` and `load_snp_subsets`, AND applied against the germline subset of the extras read to close the partial-`n_zygosity` edge case where interleaved somatic rows could mask the all-zero-normal heuristic.
- `test_scatter_show_snvs_extras` rewritten to monkey-patch `get_segment_vafs` and capture its actual input (the previous version recomputed the trend with the same expression twice and was a tautology that would not have caught a future regression where the overlays leaked into the BAF math).
- Bare `#290` everywhere (no `gh#` prefix in code prose).
- Defensive `getattr(args, "show_snvs", "het")` simplified to `args.show_snvs`.
- `# noqa: PLC0415` carries the required one-line rationale.

## Commits

1. `test: drop gh# prefix from test_export docstring (#892)` — drive-by repair of a leak from PR #1099.
2. `scatter: surface LOH and somatic SNV evidence as colored overlays (#290)` — the feature plus tests and docs.

## Test plan

- [x] `pytest test/` — 420 passed, 45 subtests passed (was 414 before; +6 new tests in this branch).
- [x] `ruff check cnvlib/ test/` — clean.
- [x] `mypy cnvlib/scatter.py cnvlib/cmdutil.py cnvlib/plots.py cnvlib/commands.py` — clean.
- [x] End-to-end smoke test renders to PNG with the expected visual: het dots gray, LOH dots reddish-purple `x` at VAF extremes, somatic dots blue `+`, external legend.
- [x] Reviewer to confirm the visual is acceptable (legend placement, color/marker choices) on a real local sample.

Fixes #290

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cnvlib/cmdutil.py
+++ b/cnvlib/cmdutil.py
@@ -46,6 +46,31 @@ def read_ga(
     return tabio.read(infile, into=GA, sample_id=sample_id, meta=meta)  # type: ignore[return-value]
 
 
+def _resolve_zygosity_freq(
+    germline_varr: VariantArray, zygosity_freq: float | None, *, warn: bool
+) -> float | None:
+    """Apply the Mutect2 all-zero-normal workaround heuristic.
+
+    When the VCF's normal-sample genotypes are all 0/0 or missing (Mutect2
+    quirk), infer zygosity from allele frequency by setting ``zygosity_freq``
+    to 0.25. Caller must pass the germline subset only (e.g. ``varr`` from a
+    ``skip_somatic=True`` read), so that the all-zero check reflects the
+    germline rows and is not masked by interleaved somatic calls.
+    """
+    if (
+        zygosity_freq is None
+        and "n_zygosity" in germline_varr
+        and not germline_varr["n_zygosity"].any()
+    ):
+        if warn:
+            logging.warning(
+                "VCF normal sample's genotypes are all 0/0 or missing; "
+                "inferring genotypes from allele frequency instead"
+            )
+        return 0.25
+    return zygosity_freq
+
+
 def load_het_snps(
     vcf_fname: str,
     sample_id: str | None = None,
@@ -64,13 +89,7 @@ def load_het_snps(
         min_depth=min_variant_depth,
         skip_somatic=True,
     )
-    if zygosity_freq is None and "n_zygosity" in varr and not varr["n_zygosity"].any():
-        # Mutect2 sets all normal genotypes to 0/0 -- work around it
-        logging.warning(
-            "VCF normal sample's genotypes are all 0/0 or missing; "
-            "inferring genotypes from allele frequency instead"
-        )
-        zygosity_freq = 0.25
+    zygosity_freq = _resolve_zygosity_freq(varr, zygosity_freq, warn=True)  # type: ignore[arg-type]
     if zygosity_freq is not None:
         varr = varr.zygosity_from_freq(zygosity_freq, 1 - zygosity_freq)  # type: ignore[attr-defined]
     if "n_zygosity" in varr:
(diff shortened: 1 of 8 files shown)


etal:
Example plots with the new elements --

Genome-wide plot, both LOH indicators and somatic SNVs:
cnvkit_b2i_genome

Chromosome-level plot, only somatic SNVs visible:
cnvkit_b2i_chr1

Issue #735, opened 2022-04-21
How to get the ensembl-gene-info.hg19.tsv file for RNAseq CNV?


xiucz:
Hi,
I see the `ensembl-gene-info.hg38.tsv` file in the `/data` , which is used for `import-rna --gene-resource `. But how to get the hg19 version, or how can I prepare for it ?

Best,
xiucz

etal:
CNVkit now ships `cnv_gene_info.py`, a console script that builds the `ensembl-gene-info.<genome>.tsv` gene-resource table for any reference genome from an Ensembl BioMart export, documented in the new "Building reference assets for other genomes" section of `doc/rna.rst` (PR #1110, current master; ships in the next release after v0.9.13).

The hg19 case is the worked example:

    cnv_gene_info.py mart_export.txt -g hg19 -o ensembl-gene-info.hg19.tsv

Columns are matched by name (case-insensitively, with common aliases), and `Gene name`, `NCBI gene ID`, and `Transcript support level` are optional.

Closing as addressed; please reopen if the generated table does not work for your input.

Issue #530, opened 2020-07-14
import-rna for non-hg38 genomes


igordot:
Are there instructions anywhere on how to generate `ensembl-gene-info.hg38.tsv` for other genome versions or other species?

Does it have to be Ensembl specifically? It looks like the first column just has to match the counts table first column. I think the only Ensembl-specific field is the transcript support level.

etal:
CNVkit now ships `cnv_gene_info.py` to build the gene-resource table for any genome from an Ensembl BioMart export, with a documentation walkthrough ("Building reference assets for other genomes", `doc/rna.rst`; PR #1110, current master; ships in the next release after v0.9.13).

This answers both of your questions: the table need not be Ensembl-specific (columns are matched by name with common aliases, and `Gene name` / `NCBI gene ID` / `Transcript support level` are optional and filled with defaults when absent), and the documentation lists the exact attributes to export for any species/assembly (mouse mm10/mm39, GRCh37, non-human, etc.).

Closing as addressed; please reopen if a non-hg38 genome still gives trouble.

Issue #479, opened 2019-11-15
Input for the import-rna command


SJRussell:
1. It's not clear what Gene Resource data to download from BioMart. I'm using the built in Hg38 gene info as a template but BioMart doesn't have "NCBI gene ID" or "Transcript support level (TSL)" available for Hg19. I'm going to try to get the original fasta files for the project and remap to Hg38, but it would be great to have a brief overview of what "Gene-resource" info is required when working with non-Hg38 genomes. 
2. I'd also like to build my own cnv expression correlates, but the input requirements for cnv_expression_correlate.py are not clear. Could you point me to a resource for building these inputs?
3. There are several [issues](https://github.com/etal/cnvkit/issues/425)/[questions ](https://www.biostars.org/p/376275/)about using counts as input for import-rna. Have the issues been fixed or should I run RSEM instead of  HTSeq-count to generate my sample input.

Thanks for your great tool and in advance for your help.

etal:
Thanks for checking in, and sorry for the trouble.
1. Offhand I'm not sure what to do about BioMart's lack of support for hg19 etc. Do you know of another BioMart source that might have these, or a way to do it in R?
2. At the moment the code is the spec, and I agree some proper docs are necessary here. (Tagging this ticket.) Some manual wrangling of the tables is needed.
3. There's a chance it's been fixed -- if it's quick to check, then try it, otherwise RSEM would be a viable workaround.

SJRussell:
Thanks for the response. For now:

1. I've mapped to hg38 instead of trying to get the right output from biomart. The difficulty was that I didn't know what fields were required for the gene-resource file.
2. If you update the documentation to include details on how to build custom expression correlates, please mention in this ticket. It seems to me that the accuracy of the algorithm depends on experiment-specific expression correlates.
3. Upon installing CNVkit with conda and running any cnvkit.py commands, I got this error:
`Traceback (most recent call last):
  File "/home/stewart/anaconda3/envs/cnvkit/bin/cnvkit.py", line 8, in <module>
    from cnvlib import commands
  File "/home/stewart/cnvkit/cnvlib/__init__.py", line 4, in <module>
    from .cmdutil import read_cna as read
  File "/home/stewart/cnvkit/cnvlib/cmdutil.py", line 7, in <module>
    from .cnary import CopyNumArray as CNA
  File "/home/stewart/cnvkit/cnvlib/cnary.py", line 9, in <module>
    from . import core, descriptives, params, smoothing
[...]

PR #1110, opened 2026-06-29
RNA: gene-info tooling for arbitrary genomes; fix bundled hg38 first-gene drop

### LINKED ISSUE #386: ensembl-gene-info file hg19
Hi,

Could you please share the hg19 version of this file for rnaseq cnv analysis or show me what columns it should have if my rsem output looks like following:

`
gene_id transcript_id.s.        length  effective_length        expected_count  TPM     FPKM
6M1-18  uc010jrh.1      473     272.4   0       0       0
7M1-2   uc003wdr.3      68      0       0       0       0

`

etal:
## Summary

Adds first-party tooling and documentation for running the RNA copy-number
pipeline on reference genomes other than the bundled hg38, and fixes a
long-standing off-by-one in the gene-info loader that silently dropped one gene
from the bundled table.

## Tooling + docs

The RNA pipeline needs a per-gene metadata table
(`data/ensembl-gene-info.hg38.tsv`), which was bundled for hg38 only with no way
to regenerate it. Users on other genomes (hg19, mouse, non-human assemblies)
had no supported path to produce the equivalent input — a question raised across
several issues over four years.

- **New `cnv_gene_info.py`** (pip console-script): normalizes an Ensembl BioMart
  TSV export into CNVkit's canonical gene-info format. Columns are matched by
  name (case-insensitive, with aliases across Ensembl releases); `--rename`
  overrides unrecognized headers and `--chromosomes` filters contigs. Output is
  validated, sorted in CNVkit's canonical chromosome order, and carries a
  leading provenance comment.
- **`doc/rna.rst`**: new section "Building reference assets for other genomes"
  (BioMart recipe, non-human `cnvkit.py access`, and the three
  correlation-table options), plus a corrected `cnv_expression_correlate.py`
  reference.
- **`doc/scripts.rst`**: documents `cnv_gene_info.py` and the previously
  undocumented `genome_instability_index.py`.

## Bug fix: gene-info loader dropped the first gene of the bundled table

`load_gene_info` read the table with `header=1`, which skips the first physical
line before the header row. That is correct only for files that carry a leading
line before the column header. The bundled `ensembl-gene-info.hg38.tsv` has its
column header on the first line, so the loader silently dropped its first data
row — `ENSG00000210049` (MT-TF, a mitochondrial tRNA) — on every hg38 RNA run.
This has been the case since the table was added (2017).

The loader now reads with `header=0, comment="#"`: the first line is the column
header, and an optional leading `#` provenance comment (as emitted by
`cnv_gene_info.py`) is ignored. All gene-info files load every gene.

### Behavior change (for release notes)

hg38 RNA analyses (`cnvkit.py import-rna`) now include one additional gene,
`ENSG00000210049` / MT-TF, when it is present in the input expression data. This
is the only change to numerical output; all other genes are unaffected.

## Testing

- `test/test_rna.py`: `BuildGeneInfoTests` (generator) and
  `LoadGeneInfoHeaderTests` (regression guard for the first-row drop).
- Full `test_rna.py` suite passes; Sphinx docs build clean.

Fixes #386, #479, #530, #735, #822.


etal:
@claude Review, please.

PR #1131, opened 2026-07-22
Fix duplicate haar segments and BAF broadcast from non-unique segment index (#1125)

### LINKED ISSUE #1125: Does haar create duplicate segments?
A user reported seeing dupes on the regression testing file, `9_2`, when re-segmenting with haar:
https://github.com/etal/cnvkit/tree/master/test/formats/regression

Is it real and reproducible, or an artifact of the user's environment?

etal:
## Summary

Re-segmenting some inputs (e.g. the regression fixture `test/formats/regression/9_2`) with `-m haar` produced pairs of segments sharing identical `(chromosome, start, end)` bounds. The root cause is a non-unique segment index that a shared post-processing step then relies on by label.

`do_segmentation` splits bins by chromosome arm and segments each arm; `segment_haar` itself calls `by_arm()` again and re-splits an arm on internal gaps, then concatenated the per-arm segment tables **without resetting the index** — each per-arm table starts at index 0, so the result carried duplicate index labels. `transfer_fields` stretches the first and last segment endpoints to the arm's bin span via `.loc[index[0]]` / `.loc[index[-1]]`; with duplicate labels, `.loc[label]` broadcasts the write to *every* row sharing that label, collapsing distinct segments onto the full-arm interval.

## Changes

**Commit 1 — duplicate haar segments**
- `segment_haar`: reset the index when concatenating per-arm tables (`ignore_index=True`) so a segmentation result always carries a unique index.
- `transfer_fields`: address the first/last segment by **position** (`.iloc`) rather than by index label, so the shared post-process is robust to any index. For a clean (unique) index this is identical to the previous behavior, so CBS and clean-index output is unchanged.

**Commit 2 — BAF broadcast on `cbs`/`none` + `--vcf`**

The same non-unique-index hazard existed on the CBS/`none` variant re-segmentation path. `variants_in_segment` returns a table indexed from 0, and concatenating two or more of them left duplicate labels; the subsequent `segarr["baf"] = variants.baf_by_ranges(segarr)` aligns by label and silently broadcast the first segment's BAF to every segment sharing that label — producing wrong per-segment BAF in the `.cns`. Fixed by resetting the index on that concat as well.

## Verification

- On the 9_2 fixture: before, 313 segments with 11 duplicate-interval pairs; after, 313 segments with 0 duplicates and 0 overlaps (segment count, log2, and probes unchanged — only the collapsed coordinates are corrected).
- Clean-index inputs (amplicon, p2-20_1/2, wgs-chr17) are byte-identical before/after.
- The BAF fix: a CBS segmentation with a copy-number step aligned to a BAF shift returns BAF `[0.5, 0.3]` after the fix versus `[0.5, 0.5]` before (second segment's true BAF silently overwritten).
- Two regression tests added (`test_segment_haar_no_duplicate_segments`, `test_cbs_variants_baf_per_segment`); both fail on the pre-fix code and pass after. `test_segmentation.py` and `test_r.py` suites green.

Fixes #1125


etal:
## CI fix: regenerated frozen regression fixtures (commit 16c4e6a)

The `test_regression.py::test_pipeline_matches_frozen` job failed because the frozen `p2-9_2.cns` / `p2-9_2.call.cns` fixtures were generated **before** this fix and encoded the corrupted output it corrects. Those fixtures are exactly the 9_2 output issue #1125 reports: the duplicate-index broadcast in `transfer_fields` had collapsed distinct haar segments onto the full-arm span, leaving 22 duplicate-interval rows and 77 overlapping intervals across the file.

Regenerated via `python test/test_regression.py --regenerate`. This is an intended numerical correction, verified as a faithful regeneration:

- Segment count unchanged (313 rows).
- `chromosome`, `log2`, and `probes` multisets are identical old vs new.
- 250 of 313 segments have unchanged `(chromosome, start, end)`; only 63 changed — precisely the segments the bug had corrupted.
- `gene`/`weight`/`depth` differ only on those 63 corrected segments (their per-segment aggregates were previously computed over the collapsed full-arm coordinate ranges, so they were wrong too).
- The `.cnr` fixture is byte-identical (`do_fix` is unaffected) and is not modified.

New fixtures have 0 duplicate and 0 overlapping intervals. Full `test_regression.py`, `test_segmentation.py`, `test_r.py`, and `test_haar.py` suites pass locally.

